PR #1, opened 2013-06-06
Avoid division by zero in posix targets

tinti:
The tick duration in microseconds may be calculated to zero depending on the
processor clock speed.

Maybe another solution could be using BogoMips for calculating the tick duration or even querying sysfs (i.e. /sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq). Both have drawbacks and in some case zero will be get as a result.

Also BogoMips has recently change and now it can be specified at build time.

### DIFF
--- a/src/os/inc/common_types.h
+++ b/src/os/inc/common_types.h
@@ -107,23 +107,45 @@
    #define OS_USED 
 #endif
 
-#if defined(_ix86_)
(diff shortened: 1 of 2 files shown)


skliper:
This change is obsolete, wrote #12 to address this issue in the context of the current code base.

PR #2, opened 2013-06-14
Use stdint as common type bases

tinti:


skliper:
The upcoming release uses common type bases.  Closing as duplicate.

PR #3, opened 2013-07-24
Pull request partial deadlock fix

tinti:
The OSAL api uses a global lock to perform some operations using mutex,
semaphores, queues and so. This leads to a possible deadlock in the program
due the following situation.

Let "I" be the timer interrupt (SIG63):

```
   |[Thread 1]
   |  ...
   |  call OS_TaskGetInfo
   |    ...
   |    lock(global)
   |    ...
   +--------------------+
                        |[Interrupt]
                        |  ...
                        |  call OS_TaskGetInfo
                        |    ...
                        |    lock(global)  // DEADLOCK
```

There are some possible solutions to avoid this problem. One is to never use any
OSAL primitives on the interrupt context, which quite limit the library. Other
would be block all signals during the global lock and let them be pending. This
approach may introduce poor performance. Other would be find out how to remove
these global locks. Another could be use safe locks on interrupt handler.

This patch implements the blocking signals solution because it is the best
trade-off of the current solutions.

### DIFF
--- a/src/os/inc/common_types.h
+++ b/src/os/inc/common_types.h
@@ -107,23 +107,45 @@
    #define OS_USED 
 #endif
 
-#if defined(_ix86_)
+#include <stdint.h>
+#include <stdbool.h>
+
+#define int8 int8_t
+#define int16 int16_t
+#define int32 int32_t
+#define int64 int64_t
+#define uint8 uint8_t
+#define uint16 uint16_t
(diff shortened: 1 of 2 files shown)


skliper:
OSAL primitives in general are not intended to be interrupt safe.  Feel free to implement and open source interrupt safe versions if you'd like, but not planning to overhaul the framework OSAL examples.

PR #4, opened 2017-01-25
OS_QueueGet bugfix

bacinchel:
When user uses OS_QueueGet  with timeout paramater, then always timeot fails even if actual data aquired


the-other-james:
This fix is incorporated in OSAL version 4.2.1a.

PR #16, opened 2019-09-16
Fix #15 unit test warnings

### LINKED ISSUE #15: Many compiler warnings in unit tests when using strict compiler flags
**Describe the bug**
For historical backward-compatibility reasons ANSI C defines string literals as `char*` type, but these strings are typically put into an `.rodata` section on modern processors/linkers, and this section is usually write-protected.  If code actually attempts to modify this value, it will generate a segmentation fault and crash.

Building with `-Wwrite-strings` causes the compiler to treat string literals as `const char *` instead, thereby generating a warning if passed or assigned to a non-const pointer value.

**To Reproduce**
Configure with `make OSAL_USER_C_FLAGS='-Wall -Werror -std=c99 -pedantic -Wwrite-strings' ENABLE_UNIT_TESTS=true` and observe many warnings generated in the unit test code.

**Expected behavior**
The build should be free of warnings, such that `-Wwrite-strings` can be enabled by default.

**System observed on:**
Ubuntu 18.04 (64-bit)

**Reporter Info**

jphickey:
**Describe the contribution**
Fixes #15 
- Remove unused variable
- Add local function prototype where missing
- Qualify all pointers to string literals as `const`

**Testing performed**
Rebuild OSAL per default configuration with unit tests enabled
Re-ran all unit tests and confirm passage

**Expected behavior changes**
No warnings generated when using `-Wall -Werror -std=c99 -pedantic -Wwrite-strings -Wstrict-prototypes'

**System(s) tested on:**
Ubuntu 18.04 (64-bit)

**Contributor Info**

### DIFF
--- a/src/unit-tests/oscore-test/ut_oscore_test.c
+++ b/src/unit-tests/oscore-test/ut_oscore_test.c
@@ -29,9 +29,9 @@ extern UT_OsLogInfo_t  g_logInfo;
 **--------------------------------------------------------------------------------*/
 
 int32  g_skipTestCase = -1;
-char*  g_skipTestCaseResult = " ";
+const char*  g_skipTestCaseResult = " ";
 
-char* g_task_names[UT_OS_TASK_LIST_LEN];
+const char* g_task_names[UT_OS_TASK_LIST_LEN];
 char  g_long_task_name[OS_MAX_API_NAME+5];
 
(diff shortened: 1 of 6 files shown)


jphickey:
Updated commit so it is based on the current `master` branch.

PR #21, opened 2019-09-18
Fix #17: Update function description blocks across OSAL

### LINKED ISSUE #17: Inline documentation on OSAL routines is incorrect/outdated in many places
**Describe the bug**
The comments in the function headers of many of the OSAL calls have become outdated and incorrect.

**To Reproduce**
N/A.  This reflects source code comments only.  Most evident on the "implementation" files which copied the comment block from the external function call.

**Expected behavior**
Comment blocks should better reflect reality of what the code does.  The "real" comment block that describes the function should reside with the function prototype.  The definition should also have a comment block but primarily for visual differentiation in the source code.  This block should not re-describe the function, as this results in multiple copies of the same description and thus could become stale/divergent again.

**Reporter Info**

jphickey:
**Describe the contribution**
Fix issue #17.

Update inline block comments in function/file headers across all of OSAL and ensure consistency with the code itself.

- All documentation covering the specific details, inputs/outputs
and return values should be on the function _declaration_, not the
definition.

- The definition should also contain a block comment for visual
distinction in the source file, but should be brief and refer
back to the prototype/declaration for detail.  Any detail information
that was indicated only with the definition is moved to the
declaration.

- Add doxygen tags/markup to external/public API block comments
and expand the descriptions of these functions where needed.

- Add block comments to the internal API prototypes, but _not_
as doxygen tags/markup.  Note these functions are internal use,
not part of the public API, and are not necessarily intended
to be stable across OSAL releases like the public API is.

- Removed stray prototypes for functions where there was no
actual definition/implementation.

**NOTE**: All changes are primarily confined to comments, except where unused/unimplemented prototypes were removed.

**Testing performed**
Rebuilt CFE/OSAL per default config, with unit test enabled.  Confirm build as successful and unit tests execute.

**Expected behavior changes**
No impact to behavior

**System(s) tested on:**
Ubuntu 18.04 (64-bit)

**Contributor Info**

### DIFF
--- a/src/os/inc/osapi-os-core.h
+++ b/src/os/inc/osapi-os-core.h
@@ -151,54 +151,94 @@ void OS_Application_Startup(void);
 ** Exported Functions
 */
 
-/*
-** Initialization of API
-*/
+/*-------------------------------------------------------------------------------------*/
+/**
+ * @brief Initialization of API
+ *
+ * Initialize the tables that the OS API uses to keep track of information
+ * about objects
+ *
+ * @returns OS_SUCCESS on success, or appropriate error code
+ */
 int32 OS_API_Init (void);
 
-/*
-** OS-specific background thread implementation - waits forever for events to occur.
-**
-** This should be called from the BSP main routine / initial thread after all other 
-** board / application initialization has taken place and all other tasks are running.
-*/
+/*-------------------------------------------------------------------------------------*/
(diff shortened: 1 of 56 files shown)


skliper:
Suggest adding OS_TimerSet note about cancelling with OS_TimerDelete.

PR #267, opened 2019-10-18
Fix #266: Rename "vxworks-ng" and "posix-ng" coverage tests

### LINKED ISSUE #266: Remove vxworks6 coverage code, rename vxworks-ng to vxworks
**Describe the bug**
For the `unit-test-coverage` code, the directory names must match those of the actual implementation they are testing.  The build scripts assume/require this name relationship.

Recently, the classic OSAL implementations were dropped and the "-ng" implementations replaced them in the main code tree, but the unit-test-coverage still has the old vxworks6 and new vxworks-ng directories.

The old `vxworks6` should be removed and the `vxworks-ng` needs to be renamed to be `vxworks`.

**To Reproduce**
Build coverage code using e.g.:
`cmake -DOSALCOVERAGE_TARGET_OSTYPE='shared;vxworks-ng' -DOSAL_INCLUDEDIR=$CONFIG_DIR $OSAL_DIR/src/unit-test-coverage`

The subsequent `make` will fail because there is no longer an OSAL implementation named `vxworks-ng`, it is named `vxworks` now.

**Expected behavior**
The build should complete successfully

**System observed on:**
Ubuntu 18.04.2 64-bit

**Reporter Info**

jphickey:
Rename to "vxworks" and "posix", respectively, to match the
actual implementation names.  The build scripts rely on
the name being the same and will fail if different.

Fixes issue #266

**Testing performed**
Build coverage test for vxworks using cmake command:

`cmake -DOSALCOVERAGE_TARGET_OSTYPE='shared;vxworks' -DOSAL_INCLUDEDIR=$CONFIG_DIR $OSAL_DIR/src/unit-test-coverage`

Followed by "make" as usual.

**Expected behavior changes**
The build completes successfully.

**System(s) tested on:**
Ubuntu 18.04.2 64-bit

**Additional context**
This only fixes the specific issue of the name mismatches and subsequent build failure.  There are a couple other issue tickets about coverage testing which this does not address.  These will be addressed in other commits/pull requests.

**Contributor Info**


jphickey:
Commit Updated to address one minor detail - the default coverage implementation if not specified on the command line was still referring to vxworks6 here.

Commit 89fec29 fixes the default to be vxworks instead.

skliper:
CCB 20191023 - reviewed and approved

PR #276, opened 2019-10-23
Fix #273: Correct format code in debug message

### LINKED ISSUE #273: Incorrect printf format code in RTEMS debug message
**Describe the bug**
Incorrect printf format code in RTEMS debug message

**To Reproduce**
Build on RTEMS with debugging messages enabled and full warnings enabled (-Wall, -Werror)
Compiler generates an error about an incorrect format:
```
osal/src/os/rtems/osfilesys.c:176:9: error: format '%lX' expects argument of type 'long unsigned int', but argument 4 has type 'char *'
```

**Expected behavior**
Code should build warning-free

**System observed on:**
Ubuntu 18.04.2 LTS (build host for i686-rtems4.11)

**Reporter Info**
Joseph Hickey, Vantage Systems Inc.

jphickey:
**Describe the contribution**
Fixes issue #273
Just adds a typecast to avoid the warning

**Testing performed**
Build code on i686-rtems4.11 platform, with -Wall -Werror and OS_DEBUG_PRINTF option enabled.

**Expected behavior changes**
Observe no more warning about the format mismatch.

**System(s) tested on:**
Ubuntu 18.04.2 LTS 64 bit (build host) using i686-rtems4.11 cross toolchain

**Contributor Info**

### DIFF
--- a/src/os/rtems/osfilesys.c
+++ b/src/os/rtems/osfilesys.c
@@ -174,7 +174,7 @@ int32 OS_FileSysStartVolume_Impl (uint32 filesys_id)
         impl->mount_fstype = RTEMS_FILESYSTEM_TYPE_RFS;
 
         OS_DEBUG("OSAL: RAM disk initialized: volume=%s device=%s address=0x%08lX\n",
-                local->volume_name, impl->blockdev_name, local->address);
+                local->volume_name, impl->blockdev_name, (unsigned long)local->address);
 
         return_code = OS_SUCCESS;
         break;


jphickey:
> Would it make sense to report as 64 bit, since address could be up to this size?

As it is `unsigned long` it will typically be 64 bits on a 64-bit machine anyway.  The only (minor) issue is you will not get a bunch of leading zeros.  But it should work just fine on a 64-bit machine when that day comes.

skliper:
> > Would it make sense to report as 64 bit, since address could be up to this size?
> 
> As it is `unsigned long` it will typically be 64 bits on a 64-bit machine anyway. The only (minor) issue is you will not get a bunch of leading zeros. But it should work just fine on a 64-bit machine when that day comes.

Isn't the format of 0x%08lX only up to 8 digits?

jphickey:
> Isn't the format of 0x%08lX only up to 8 digits?

The printf field width (8) is a minimum width.  Printf formatters will go beyond the specified width to print the value; they don't truncate, they just expand.  So on a 64-bit platform it will just print more digits as needed, it'll work fine.

skliper:

PR #277, opened 2019-10-23
Fix #271: Correct interval time computations

### LINKED ISSUE #271: Incorrect time to first callback when using OS_TimerCreate API
**Describe the bug**
There discrepancies between the expected first callback and the time the first callback actually occurs, for timers created via `OS_TimerCreate()`.  This is with regards to the "start_time" parameter.

**To Reproduce**
The `timer-test` example program exposes this bug.  It was never seen because on POSIX the difference is not substantial enough to cause failure, but on VxWorks it does fail.

General sequence is:
Create a timer via `OS_TimerCreate()`
Set the timer using `OS_TimerSet()` with start_time=2000000 (2 sec) and interval_time=400000 (400ms).

The time between the OS_TimerSet call and the first timer callback should be 2 seconds (within system timer tick tolerances) but it is approximately 2.4 seconds on POSIX.  On VxWorks the difference is even more substantial, causing failure of the timer-test example program.

**Expected behavior**
The time between the OS_TimerSet call and the first timer callback should be equivalent to the `start_time` parameter from the OS_TimerSet call.

**System observed on:**
Ubuntu 18.04.2 LTS, 64 bit, using `posix` OSAL build
VxWorks 6.9 on MCP750, using `vxworks` OSAL build
Both running the "timer-test" example program.

**Additional context**
This appears to be limited to cases where the backward-compatible OS_TimerCreate API is used.  In this case an implicit timebase object is created just to service the single timer.  In this mode, there is an initial condition issue in the wait routine, and the first interval therefore is not correct. 

If the timer callback is registered via `OS_TimeBaseCreate()` followed by `OS_TimerAdd()`, then it is OK.

**Reporter Info**

jphickey:
**Describe the contribution**
Fixes issue #271, corrects the time intervals returned by the sigwait() routines for all three operating systems (POSIX, RTEMS, VxWorks).

**Testing performed**
Testing is focused on the operation of the `timer-test` example program and variations thereof.
Verified that all callbacks from all timers correlate with the configured "start_time" and "interval_time" parameters passed to the OS_TimerSet() configuration.

Verified CFE still boots and runs OK. (stops FLYWHEEL as normal based on 1Hz tick)

**Expected behavior changes**
No changes to applications/CFE.

The `timer-test` example program now runs and passes consistently on all platforms.

**System(s) tested on:**
Ubuntu 18.04.2 LTS 64 bit (simulation)
RTEMS 4.11 pc686 BSP running in QEMU
MPC750 VxWorks 6.9

**Contributor Info**

### DIFF
--- a/src/os/posix/ostimer.c
+++ b/src/os/posix/ostimer.c
@@ -135,7 +135,7 @@ void OS_TimeBaseUnlock_Impl(uint32 local_id)
  *  Purpose: Local helper routine, not part of OSAL API.
  *
  *-----------------------------------------------------------------*/
-static uint32 OS_TimeBase_SoftWaitImpl(uint32 timer_id)
+static uint32 OS_TimeBase_SigWaitImpl(uint32 timer_id)
 {
     int ret;
     OS_impl_timebase_internal_record_t *local;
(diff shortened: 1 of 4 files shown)


jphickey:
> Any need in posix/linux to check for a timer interval that is different due to the tick rate (like RTEMS and vxWorks?)

On all my testing on various Linux/Glibc implementations, the timers have always worked correctly.  That is, for any interval you specify, the implementation generates a signal at that interval, regardless of the system timer tick period.  I haven't dug through the Glibc implementation, but it seems they are implementing whatever special sauce is needed to obtain the specific interval that was requested.  I guess that's one of the benefits of using a full-fat C library instead of a trimmed-down one.

> In vxWorks was the posix timer API used so it could be similar to the posix code? Does the vxWorks API offer any features that the posix does not?

Does VxWorks offer a better timer API than the posix(-ish) API that is being used here?  This timer setup code was lifted from the vxworks6 OSAL and the same API was used there.

Honestly, I was surprised to see such a blatant difference between the requested interval and the actual provided interval, particularly because the requested interval was an exact multiple of timer ticks, yet timer_settime still added an extra tick to it, so it was basically just wrong at that point.  It strikes me as a bug in the implementation.

Reading back the interval with timer_gettime is basically just a workaround for this bug, but should work fine on platforms that work correctly too.

skliper:

PR #278, opened 2019-10-23
Fix #274: Avoid possible deadlock of timer callback

### LINKED ISSUE #274: Possible deadlock of timer callbacks
**Describe the bug**
If the actual timer interval is larger than the requested timer interval, due to the system tick timer resolution, then it is possible for timers to become deadlocked.

**To Reproduce**
Configure a timer with `OS_TimerCreate()` and `OS_TimerSet()`.  Configure the interval such that it must be rounded up to a whole number of system ticks.

The timer works fine for a while, but then callbacks will cease once the first overrun occurs.

**Expected behavior**
Callbacks should continue until canceled by the application.

**System observed on:**
RTEMS 4.11 (pc686) running in QEMU
MPC750 VxWorks 6.9

**Additional context**
Discovered as part of the fix for #271, submitted as a separate bug for review as it affects the shared/common layer.

**Reporter Info**

jphickey:
**Describe the contribution**
Fix issue #274, avoids timer callback deadlock if actual tick_time is consistently bigger than the configured iinterval_time.

**Testing performed**
Modify `timer-test` to choose an interval that is not a multiple of the system clock tick, such that it will be rounded up to the next tick by the implementation.

Verify that evenually, two callbacks are generated in a single time tick, such that the long-term average of callbacks over time is consistent with the configured interval_time.

Confirm that callbacks continue normally until the application is stopped, as expected.

**Expected behavior changes**
Callbacks no longer cease once the "overrun" condition occurs (callbacks behind by >=1 full interval time)

Confirm consistent behavior on POSIX, RTEMS, and VxWorks

**System(s) tested on:**
Ubuntu 18.04.2 LTS 64-bit (POSIX/simulation)
RTEMS 4.11 pc686 BSP running in QEMU
MPC750 VxWorks 6.9

**Contributor Info**

### DIFF
--- a/src/os/shared/osapi-timebase.c
+++ b/src/os/shared/osapi-timebase.c
@@ -494,7 +494,7 @@ void OS_TimeBase_CallbackThread(uint32 timebase_id)
                 timecb = &OS_timecb_table[curr_cb_local_id];
                 saved_wait_time = timecb->wait_time;
                 timecb->wait_time -= tick_time;
-                if (timecb->wait_time <= 0)
+                while (timecb->wait_time <= 0)
                 {
                     timecb->wait_time += timecb->interval_time;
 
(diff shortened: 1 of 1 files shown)


skliper:

PR #281, opened 2019-10-31
Integration Candidate 20191030

### LINKED ISSUE #230: Improve coverage test for "shared" layer
This is a follow-on to original ticket #28

As part of the original change sets, a basic coverage test was added for the shared layer.  It calls most functions in a nominal case, but does not yet cover a sufficient number of the error cases.  For instance the module loader only has 72% of line coverage.  Others are between 80-90% but could still be improved (files, queues, timers).

It has been discussed in CCB meetings during the review for #28 that we should make another pass through the coverage test logs to improve the error case coverage prior to releasing the next OSAL.

### LINKED ISSUE #266: Remove vxworks6 coverage code, rename vxworks-ng to vxworks

### LINKED ISSUE #269: Incorrect test of taskSpawn result in VxWorks implementation
**Describe the bug**
The VxWorks implementation is testing for an error result from the `taskSpawn` API call by comparing with `(TASK_ID)0`.  This is incorrect, the VxWorks API documentation says that when taskSpawn fails, it returns the value `ERROR`, not 0.

**To Reproduce**
This can only be reproduced in unit test.

**Expected behavior**
The error handing path should be taken when `taskSpawn` returns `ERROR`.

**System observed on:**
Ubuntu 18.04.2 LTS 64 bit (running OSAL coverage unit tests)

**Additional context**
Discovered when implementing the coverage error path unit tests in #230 

**Reporter Info**

### LINKED ISSUE #271: Incorrect time to first callback when using OS_TimerCreate API

### LINKED ISSUE #272: sem-speed-test program creates tasks with zero stack size
**Describe the bug**
For RTEMS and POSIX this is not an issue because these implementations will just use a minimum stack size, but VxWorks actually creates the task with a very small stack.  The worker tasks are then likely to overrun the stack, causing undefined behavior.

**To Reproduce**
Execute sem-speed-test on the MPC750 test platform
Observe inconsistent behavior during/after test (sometimes it completes OK, sometimes errors, occasionally even a reboot).

**Expected behavior**
The test should complete successfully.

**System observed on:**
MCP750, VxWorks 6.9

**Reporter Info**

### LINKED ISSUE #273: Incorrect printf format code in RTEMS debug message

### LINKED ISSUE #274: Possible deadlock of timer callbacks

skliper:
**Describe the contribution**
Fixes #230, fixes #266, fixes #269
Fixes #271, fixes #272, fixes #273, fixes #274

**Testing performed**
1. Checked out bundle with OSAL and cFE ic-20191030 branches 
1. make ENABLE_UNIT_TESTS=TRUE SIMULATION=native prep
1. make
1. make install
1. make test
Built without warnings, all tests passed except osal_timer_UT (nominal result on linux)
executed cfe, successful startup with no warnings

**Expected behavior changes**
Coverage test updates (more coverage and works for vxworks)
Timer bug fixes
sem-speed-test functional test fix for vxworks

**System(s) tested on:**
 - cFS dev server
 - OS: Ubuntu 16.04
 - Versions: bundle with OSAL and cFE ic-20191030 branches

**Additional context**
None

**Contributor Info**
Jacob Hageman - NASA/GSFC


skliper:
CCB 20191106 - approved for merge

Issue #308, opened 2019-12-12
Cppcheck for os dir. 


avan989:
**Is your feature request related to a problem? Please describe.**
From issue #303 

```
[osal/src/os/portable/os-impl-posix-dl.c:54]: (style) struct member 'OS_impl_module_internal_record_t::dl_handle' is never used.
[osal/src/os/posix/os-posix.h:110] -> [osal/src/os/posix/osapi.c:737]: (style, inconclusive) Function 'OS_Posix_InternalTaskCreate_Impl' argument 1 names different: declaration 'thr' definition 'pthr'.
[osal/src/os/posix/os-posix.h:110] -> [osal/src/os/posix/osapi.c:737]: (style, inconclusive) Function 'OS_Posix_InternalTaskCreate_Impl' argument 4 names different: declaration 'Entry' definition 'entry'.
[osal/src/os/posix/osapi.c:163] -> [osal/src/os/posix/osapi.c:2264]: (style, inconclusive) Function 'OS_CompAbsDelayTime' argument 1 names different: declaration 'milli_second' definition 'msecs'.
[osal/src/os/rtems/osloader.c:43]: (style) struct member 'OS_impl_module_internal_record_t::dl_handle' is never used.
```

**Requester Info**
Anh Van, NASA Goddard

PR #310, opened 2019-12-12
issue #308, Resolve cpp warning in os dir.

avan989:
**Describe the contribution**
Resolve cppcheck for os dir.  Possible cppcheck bug for structure member - suppress. 

**Testing performed**
1. make prep, make, make install
2. run 

**System(s) tested on:**
 - Hardware
 - Ubuntu 18.04
-  CFE 6.6

**Contributor Info**
Anh Van, NASA Goddard

### DIFF
--- a/src/os/portable/os-impl-posix-dl.c
+++ b/src/os/portable/os-impl-posix-dl.c
@@ -51,6 +51,7 @@
  */
 typedef struct
 {
+    /* cppcheck-suppress unusedStructMember */
     void *dl_handle;
 } OS_impl_module_internal_record_t;
 
--- a/src/os/posix/os-posix.h
+++ b/src/os/posix/os-posix.h
@@ -107,7 +107,7 @@ int32 OS_Posix_StreamAPI_Impl_Init(void);
 int32 OS_Posix_DirAPI_Impl_Init(void);
 int32 OS_Posix_FileSysAPI_Impl_Init(void);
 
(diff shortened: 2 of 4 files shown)


src/os/portable/os-impl-posix-dl.c
```diff
@@ -51,6 +51,7 @@
  */
 typedef struct
 {
+    /* cppcheck-suppress unusedStructMember */
```
jphickey: This seems suspicious because the dl_handle is absolutely used, but the code is conditional upon `OS_MAX_MODULES > 0` in the config.

Maybe cppcheck is complaining because it doesn't know what OS_MAX_MODULES is set to, therefore assumes that the conditional is false?

skliper: Right.  Cppcheck is checking both conditions.  I prefer this specific suppression vs passing in OS_MAX_MODULES to cppcheck, in that I don't really care about this specific warning but if future ones show up for that condition being false I want to know.  

Alternatively, if the condition really can never happen, then it probably shouldn't be there (and add a verify that OS_MAX_MODULES > 0).

jphickey: The condition of OS_MAX_MODULES being 0 _can_ happen.  But in that case the structure isn't instantiated at all. cppcheck just does not know that.

I think a better long-term solution would be to work this into the cmake-based osconfig options (to replace the preprocessor-based logic).  This will get rid of these types of `#if` conditionals in here and probably clear up this warning too.

skliper:

PR #317, opened 2019-12-13
Exception/Interrupt API updates

jphickey:
**Describe the contribution**
Fix #316

- Ensure the FPU functions that are actually not implemented in RTEMS/POSIX all return OS_ERR_NOT_IMPLEMENTED, not OS_SUCCESS.
- The RTEMS IntEnable/Disable API should not be a duplicate of the IntLock/Unlock API, as the semantics are slightly different.
  The Enable/Disable API should return OS_ERR_NOT_IMPLEMENTED.
- The OS_FPUExcGetMask API should confirm that the output buffer is not NULL.

**Testing performed**
Sanity check on basic CFE operation on native/simulation build - no changes to FSW behavior.

Modified APIs are exercised in the OSAL unit tests.  Unit tests executed on RTEMS, POSIX (native), and VxWorks in conjunction with changeset for #313.

**Expected behavior changes**
No change to FSW.
Unit tests related to these particular API calls now all PASS or are skipped appropriately for the ones that are actually not implemented.

**System(s) tested on:**
- Ubuntu 18.04 LTS 64-bit (native/posix)
- RTEMS 4.11 on i686/pc (QEMU-emulated target)
- VxWorks 6.9 on PPC/MCP750

**Contributor Info**
Joseph Hickey, Vantage Systems, Inc.

**Community contributors**
You must attach a signed CLA (required for acceptance) or reference one already submitted

PR #318, opened 2019-12-13
Use UT assert for OSAL unit tests

### LINKED ISSUE #313: Code in "unit-test" should use UT Assert
**Describe the bug**
The tests in the "unit test" directory use an entirely different result reporting mechanism.  Although they do currently link with UT assert, they keep a local record of test cases and do not use any of the UT assert functions.

This is a maintenance issue.  The code makes extensive use of macros which makes it difficult to understand and also duplicates much of the logic that UT assert already has.  Furthermore, the tests do not run on all OSAL platforms (e.g. RTEMS) because the macros are incomplete, even though UT assert does support this platform.

**To Reproduce**
Build with ENABLE_UNIT_TESTS=TRUE and run all tests.  The osal "unit-tests" produce entirely different log files than the rest of the tests.

**Expected behavior**
The osal "unit-tests" should be more consistent in behavior and log file format with all the other tests.

**System observed on:**
 - Ubuntu 18.04 LTS 64-bit

**Additional context**
This is becoming more of an issue due to a parallel effort to avoid clobbering `CMAKE_C_FLAGS` (#312).  But this unit test build currently relies on `CMAKE_C_FLAGS` and OS-specific definitions (e.g. `_LINUX_OS_`, `_VXWORKS_OS_`, etc) to pick the right macros to use.  So this breaks the unit test if not setting `CMAKE_C_FLAGS` anymore.

Although it could probably be patched up again with some more hacks, it is probably worth some extra effort to just clean this up properly so it stops being a perpetual issue.

Since most things are macro based already, it is likely that simply changing the macro definitions to direct the output to UT assert instead of the local structure is probably sufficient.

**Reporter Info**

jphickey:
**Describe the contribution**

Fix #313

Refactor the "unit-tests" (extended/functional tests) for OSAL to use the UT assert library for test case reporting and platform support.

Reduce or eliminate the use of macros for platform abstraction wherever possible.  This removes most platform-specific logic from the test cases, leaving that to the OSAL/BSP abstraction.

**Testing performed**
Build unit tests for all three supported test platforms and confirm operation (POSIX, VxWorks, RTEMS).

**Expected behavior changes**
Tests were previously only being built/executed on POSIX.  Now the tests are built on all OSAL platforms, using the UT assert library.

Output logfiles are in provided by UT assert and are therefore now in the same UT assert style.  This test set no longer has a different set of output files.

**System(s) tested on:**
 - Ubuntu 18.04 LTS 64-bit (native/posix)
 - RTEMS 4.11 on i686 (using QEMU target)
 - VxWorks 6.9 on PPC/MCP750 

**Additional context**
This should be pulled in parallel with #317 to ensure that the POSIX platform continues to  pass all tests.  Note that the relevant tests (FPU, interrupt) were being skipped on POSIX., but with this change all tests are executed on all platforms.

**Contributor Info**

**Community contributors**


src/unit-tests/osloader-test/ut_module.c
```diff
@@ -3,6 +3,7 @@ void MODULE_NAME(void)
   volatile int i;
   i = 1; 
   i = i + 1;  /* squelch set-but-not-used compiler warning */
+  (void)i;
```
skliper: What's this for?

jphickey: It just quiets down a warning from cppcheck that "i" was set but not used.  The statement before it (the increment) worked to avoid a gcc warning about the same but cppcheck still complained.

Note that as part of the testing I did run this through cppcheck and resolve any remaining issues (should have noted that in the pull request)

skliper:

PR #319, opened 2019-12-13
Refactor OSAL C library stubs

### LINKED ISSUE #297: Alternative approach to compile-time function substitutions for UT
Currently, the OSAL coverage test overrides C library calls using a combination of a single "override" header file that correlates with the C library header file of the same name.  This file declares stub versions of the same functions but with an `OCS_` prefix.  

This is combined with a "stub-map-to-real.h" header that uses `#define` statements to divert the original calls to the substitute (OCS) version at compile time.

The C library calls need to be stubbed out in this manner, because they cannot be done at link time (as the UT still needs to link with the real C library, unlike higher-level libraries).

Put the `#define` statements in the substitute header itself, rather than in a local map file, and move the `OCS_` prototypes and declarations to a new, separate header file.  

The existing method of using a separate map file to provide the substitutions, but this complicates the build of the source units under test as this file must be explicitly included somehow.

**Additional context**
The change alleviates the need to inject the `stup-map-to-real.h` file as part of the  compilation of the source unit under test, as it gets the overrides implicitly though the include path instead, so it simplifies the build for UT that needs this feature.

The trade-off is that it requires a _pair_ of override headers for each real header being overridden, i.e. one for the `OCS_` replacements and one for the actual `#define` statements to do the mapping.  So for the OSAL coverage test, which override headers for about 80 different system headers, this adds quite a few files.

**Requester Info**

jphickey:
**Describe the contribution**

Fix #297

Rather than utilizing a separate "stub-map-to-real.h" file that would need to be injected into the build of the unit under test, this builds the mappings into the mapping into the override files themselves.

The stub functions are then moved into a separate header under a unique name with an OCS_ prefix.

This introduces a number of new files, as what previously required a single header now requires two, but it it eliminates the separate map file.

**Testing performed**
Build coverage test before and after change, confirm no changes to test operation or result.

**Expected behavior changes**
None, just a structural change to the source file layout/patterns

**System(s) tested on:**
- Ubuntu 18.04 LTS, 64 bit

**Contributor Info**

**Community contributors**


skliper:

skliper:
Merge pending copyright header in new files.

PR #325, opened 2019-12-18
Use alternate setup function for Tests

### LINKED ISSUE #294: Alternative entry point for UT-Assert based test applications
**Describe the bug**
Applications which utilize the UT Assert library to do testing (either functional or coverage) share the same entry point with "normal" OSAL applications, specifically the  `OS_Application_Startup` function.

For testing, this is not ideal as there is some extra setup/teardown when using UT Assert.  This prevents using the OSAL-provided BSP directly and necessitates an alternative BSP that includes the extra setup/teardown.

If the UT Assert based applications used a differently-named entry point, this could be properly layered and some duplicate logic in the OSAL BSPs can then be removed.

**To Reproduce**
Build with ENABLE_UNIT_TESTS=TRUE, and observe that the system builds a ut-bsp in addition to the normal OSAL BSP.  On pc-linux for instance, both of this libraries include a `main()` function.  The difference is that the ut-bsp includes extra UT Assert setup/teardown logic before and after the call to OS_Application_Startup().  It also duplicates the `OS_VolumeTable` object.

**Expected behavior**
The `main()` routine and `OS_VolumeTable` should _not_ be duplicated.  Rather, the UT Assert setup/teardown should be an extension layer to the normal BSP.

**System observed on:**
Ubuntu 18.04 LTS, 64 Bit

**Reporter Info**

jphickey:
**Describe the contribution**

Fix #294

Implements a more layered architecture for test code. Rather than having each test provide `OS_Application_Startup()` directly, now the UT assert library provides `OS_Application_Startup()` in common code.

The actual test setup is now done by a function called UtTest_Setup().

This also introduces a new function `OS_Application_Run()` to implement the test routine.  This replaces the existing `UtTest_Run()` function.  

Note - by using a more generic name the `OS_Application_Run()` routine can apply to both normal applications and test applications.  This is a prerequisite to a future removal of the separate UT BSP.

This allows for better layering where UT-Assert provides the notion of an "OSAL application".  This is a step toward simplifying the BSP architecture, eventually combining the normal application
and unit test BSP code into a single BSP.

**Testing performed**
Build with default config and ENABLE_UNIT_TESTS=TRUE for all three supported platforms (pc-linux, pc-rtems, and mcp750-vxworks).  

Confirm normal operation of CFE core (no change) and all unit tests run normally.

**Expected behavior changes**
No changes to behavior.

**System(s) tested on:**
- Ubuntu 18.04 LTS 64-bit (via SIMULATION=native)

**Additional context**
Needs to be merged in coordination with issue in nasa/cfe#435

**Contributor Info**

**Community contributors**


skliper:
CCB 20190108 - Reviewed and approved for integration candidate

PR #336, opened 2019-12-31
Clear pending signal when configuring timer

### LINKED ISSUE #335: osal_timer_UT periodic failure
**Describe the bug**
osal_timer_UT periodically reports a failure on linux for the nominal OS_TimerSet case

**To Reproduce**
1. Build and run
```
make ENABLE_UNIT_TESTS=TRUE SIMULATION=native prep
make
make install
build/exe/cpu1/osal_timer_UT
```

**Expected behavior**
Should pass

**Code snips**
See ut_ostimer_test.c for UT_os_timercallback (the timer callback function)
See ut_ostimer_timerio_test.c line 654 for the failure location

**System observed on:**
 - cFS Dev 2 Server
 - OS: Ubuntu 18.04
 - Versions: happened to be running the ic-20191230 branches, but not unique (historical issue)

**Additional context**
Added after the tolerance check in the callback function:
```
printf("\ndeltaTime = %d\n", deltaTime);
printf("g_toleranceVal = %d\n", g_toleranceVal);
printf("currIntervalTime = %d\n", currIntervalTime);
printf("prevIntervalTime = %d\n", prevIntervalTime);
printf("endTime = %d sec, %d msec\n", endTime.seconds, endTime.microsecs);
printf("currTime = %d sec, %d msec\n", currTime.seconds, currTime.microsecs);
```
failure result was seen due to unexpected first interval of 1056, where interval was set to 500000:
```
deltaTime = 0
g_toleranceVal = 25000
currIntervalTime = 0
prevIntervalTime = 0
endTime = 1577812468 sec, 179711 msec
currTime = 1577812468 sec, 179711 msec

deltaTime = 1056
g_toleranceVal = 25000
currIntervalTime = 1056
prevIntervalTime = 0
endTime = 1577812468 sec, 180767 msec
currTime = 1577812468 sec, 179711 msec

deltaTime = 498945
g_toleranceVal = 25000
currIntervalTime = 500001
prevIntervalTime = 1056
endTime = 1577812468 sec, 680768 msec
currTime = 1577812468 sec, 180767 msec

deltaTime = 12
g_toleranceVal = 25000
currIntervalTime = 499989
prevIntervalTime = 500001
endTime = 1577812469 sec, 180757 msec
currTime = 1577812468 sec, 680768 msec
```
**Reporter Info**
Jacob Hageman - NASA/GSFC

jphickey:
**Describe the contribution**

Fix #335 

On POSIX, call sigtimedwait() on the selected RT signal as part of the set up for the timebase.  This ensures that if the signal is already pending, it will be cleared.

This also simplifies the timer callback routine in the UT code, cleaning up some unnecessary extra logic.

**Testing performed**
Build with ENABLE_UNIT_TESTS=TRUE and execute `osal_timer_UT` in a repeated loop.

Prior to this fix, it would typically fail after about 10-20 iterations, sometimes fewer.
After this fix, it does not fail (100+ iterations, no failures).

Confirmed normal operation of CFE core (no change).

**Expected behavior changes**
Fixes occasional failures in the nominal timer test.
No change to FSW code

**System(s) tested on:**
Ubuntu 18.04 LTS, 64 Bit

**Additional context**
Tracked down the root cause of the occasional timer failures to the timer test that preceded it.  The previous test involved a short interval (5us).  Depending on the timing of the preceding delete call, the timer signal might fire again before the timer is fully un-configured (i.e. during the delete process).  

In this case, the system will be left with the signal still pending (blocked) but with no task running to accept/clear it.

This is OK until the next timer is configured, and the same signal will be selected (the first RT signal).  In this case, because the signal was already pending from the previous config, it results in an extra "spurious" initial callback at the start.  This in turn was interfering with the difference calculations in the timer UT.

It is unlikely that this issue would be seen in FSW code, as it depends on deleting and recreating timers and FSW generally does not do this (i.e. it sets up timers once).

The main part of the fix is to call `sigtimedwait()` with a zero timeout to ensure that the signal is not pending.

Additionally, there is some cleanup in the UT callback code, which fixes two other possibilities.  These were not occurring, but potentially could.
- The interval computation could be incorrect if the time elapsed was greater than 1 second.
- Always set the currTime/endTime in case there is a lag time in stopping the timer.  But only set "g_status" once.
- Remove unneeded/invalid "OS_TimerSet" call.

**Contributor Info**

**Community contributors**

### DIFF
--- a/src/os/posix/ostimer.c
+++ b/src/os/posix/ostimer.c
@@ -325,6 +325,7 @@ int32 OS_TimeBaseCreate_Impl(uint32 timer_id)
     int    status;
     int    i;
     struct sigevent   evp;
+    struct timespec ts;
     OS_impl_timebase_internal_record_t *local;
     OS_common_record_t *global;
     OS_U32ValueWrapper_t arg;
@@ -402,6 +403,34 @@ int32 OS_TimeBaseCreate_Impl(uint32 timer_id)
 
             sigemptyset(&local->sigset);
             sigaddset(&local->sigset, local->assigned_signal);
+
+            /*
+             * Ensure that the chosen signal is NOT already pending.
+             *
+             * Perform a "sigtimedwait" with a zero timeout to poll the
+             * status of the selected signal.  RT signals are also queued,
+             * so this needs to be called in a loop to until sigtimedwait()
+             * returns an error.
+             *
+             * The max number of signals that can be queued is available
+             * via sysconf() as the _SC_SIGQUEUE_MAX value.
+             *
+             * The output is irrelevant here; the objective is to just ensure
+             * that the signal is not already pending.
+             */
+            i = sysconf( _SC_SIGQUEUE_MAX);
+            do
+            {
+                ts.tv_sec = 0;
+                ts.tv_nsec = 0;
+                if (sigtimedwait(&local->sigset, NULL, &ts) < 0)
+                {
+                    /* signal is NOT pending */
+                    break;
+                }
+                --i;
+            }
+            while(i > 0);
 
             /*
             **  Initialize the sigevent structures for the handler.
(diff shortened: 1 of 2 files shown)


skliper:
CCB 20190108 - Reviewed and approved for integration candidate

PR #360, opened 2020-01-22
Fix #359: src/os/shared/osapi-module.c: handle return value of OS_ModuleLoad_Impl

### LINKED ISSUE #359: os/shared/osapi-module.c: the return value of the call to OS_ModuleLoad_Impl should be checked

I have noticed this issue while working on making the OSAL's tests pass on macOS (see #352). I am reporting this issue separately from the rest of the changeset suggested there.

The issue is in the following fragment of code as found in the current master branch (as of https://github.com/nasa/osal/commit/64ad0f551c01ab846bc79b23411e72cddf987dde):

```
    if(return_code == OS_SUCCESS)
    {
        /*
         * Save all the data to our own internal module table, but
         * only if OS_MAX_MODULES is greater than 0
         */
#if (OS_MAX_MODULES > 0)
        memset(&OS_module_table[local_id], 0, sizeof(OS_module_internal_record_t));
        strncpy(OS_module_table[local_id].module_name, module_name, OS_MAX_API_NAME);
        strncpy(OS_module_table[local_id].file_name, filename, OS_MAX_PATH_LEN);
        record->name_entry = OS_module_table[local_id].module_name;

        /* Now call the OS-specific implementation.  This reads info from the module table. */
        return_code = OS_ModuleLoad_Impl(local_id, translated_path);
#endif

        /* Check result, finalize record, and unlock global table. */
        return_code = OS_ObjectIdFinalizeNew(return_code, record, module_id);
#endif
```

We can see that the return value of the `OS_ModuleLoad_Impl` is not tested for `!= OS_SUCCESS` so the execution can continue even if the call to this function returns an error.

This is a direct link to the line: https://github.com/nasa/osal/blob/64ad0f551c01ab846bc79b23411e72cddf987dde/src/os/shared/osapi-module.c#L244

Most of the code in OSAL does check the return values: `if(return_code == OS_SUCCESS)` or `if(return_code != OS_SUCCESS)` so it makes sense to create such a check in this case too.

I am going to create a Pull Request that addresses this issue.

To the best of my knowledge,  a proper return value checking is the only way to get fast feedback from a call to the `OS_ModuleLoad_Impl` if something goes wrong.

**Additional context**

None.

**Requester Info**

Stanislav Pankevich, independent contribution.

stanislaw:
**Describe the contribution**

This changeset fixes #359.

**Testing performed**

The change is not detectable by the current test suite in the current master branch (as of 64ad0f551c01ab846bc79b23411e72cddf987dde) but I can confirm that building OSAL and running its tests on Ubuntu Linux and macOS works fine.

**Expected behavior changes**

This changeset only adds a handling of the `OS_ModuleLoad_Impl` function's return value. In all nominal cases there should be no changes in the existing behavior.

**System(s) tested on**
 - Hardware: Linux laptop, macOS laptop
 - OS: Ubuntu 19.04, macOS 10.14.6
 - Versions: OSAL as of 64ad0f551c01ab846bc79b23411e72cddf987dde

**Additional context**
None.

**Third party code**
None.

Stanislav Pankevich, personal

The signed individual CLA has been sent today to the email specified in the CLA document.

### DIFF
--- a/src/os/shared/osapi-module.c
+++ b/src/os/shared/osapi-module.c
@@ -242,6 +242,10 @@ int32 OS_ModuleLoad ( uint32 *module_id, const char *module_name, const char *fi
 
         /* Now call the OS-specific implementation.  This reads info from the module table. */
         return_code = OS_ModuleLoad_Impl(local_id, translated_path);
+        if (return_code != OS_SUCCESS)
+        {
+            return return_code;
+        }
 #endif
 
         /* Check result, finalize record, and unlock global table. */


skliper:
Suggest following established pattern - 
```
if(return_code == OS_SUCCESS)
{
  return_code = new call
}
```

Basically separate out the #ifdef'ed section as it's own check following this pattern (still surrounded by #ifdef), then OS_ObjectIdFinalizeNew should also follow the pattern above.

skliper:
Actually, slight change... since you need to unlock even if the load fails

> Suggest following established pattern -
> 
> ```
> if(return_code == OS_SUCCESS)
> {
>   return_code = new call
> }
> ```
> 
> Basically separate out the #ifdef'ed section as it's own check following this pattern (still surrounded by #ifdef), then OS_ObjectIdFinalizeNew should also follow the pattern above.

Actually retract that.  Needs refactor since you want to unlock if locked.

jphickey:
Needs further description as I'm not sure there is a real issue here.  The return code is passed through `OS_ObjectIdFinalizeNew()` which handles both success (complete allocation) and failure (clear allocation) and unlocks the global table - which is why it needs to be called always in both success and failure cases.

stanislaw:
Thanks for the insight @jphickey. Now I understand the code very well and of course my patch is wrong. Sorry about not considering this code and its surroundings in more detail.

I would still like to mention that I have never seen this kind of error handling before: passing a return result down to a next function and doing some actions depending on this argument. This is a new pattern to me and now I see that it is used consistently in many places in the OSAL code. The only disadvantage I see about this pattern is that an error handling is postponed for a while and pushed one level down so you lose the visibility of the control flow on the level of the function whose return value is not handled immediately. Additionally passing of a return value of a function through another function is I think something I see for the first time. 

Just to mention a possible alternative, this is how I could see this code be written in a bit more explicit way. The `OS_ObjectIdFinalizeNew` could be split into two functions: `OS_ObjectIdFinalizeNew` and `OS_ObjectIdClearFailed` which would each handle only one branch: success or failure. Both functions would not have the return code argument in their signatures. Then the code would look like this:

```c
      /* Now call the OS-specific implementation.  This reads info from the table. */
      return_code = OS_CountSemCreate_Impl(local_id, sem_initial_value, options);
      if (return_code != OS_SUCCESS) {
        return OS_ObjectIdClearFailed(record, sem_id);
      }
      return_code = OS_ObjectIdFinalizeNew(record, sem_id);
```

This is just an alternative way to approach this code and I am not proposing any change like this because the existing solution is very clear now. I am closing both the issue and the PR.

jphickey:
The idea is to keep the behavior as consistent as possible, and adhere to the DRY principle.  The clean-up after a new ID allocation always includes exporting the ID value and unlocking, which need to be done for ALL cases.  Splitting it into two functions would mean that this logic is repeated between the two, and also logic is repeated in every use case  (the `if`).  This repetition is not needed and just provides opportunities for inconsistency.

PR #372, opened 2020-03-19
Fix #349, OS_SocketAccept() fix

### LINKED ISSUE #349: OS_SocketAccept fails
**Describe the bug**
When using OS_SocketAccept on a stream (TCP) server socket, it fails because the accept code calls OS_ObjectIdGetById() passing a OS_LOCK_MODE_REFCOUNT, but then OS_SocketAccept sees the refcount > 0 and fails out.

**To Reproduce**
Create code (e.g. an app) that creates a TCP socket server and try using OS_SocketAccept() to receive the incoming connections.

**Expected behavior**
OS_SocketAccept should work as expected/documented.

**Code snips**
OS_SockAddr_t Addr;
int Socket = 0, ClientFd;
int32 Status = 0;

OS_SocketAddrInit(&Addr, OS_SocketDomain_INET);
OS_SocketAddrFromString(Addr, "127.0.0.1");
OS_SocketOpen(&Socket, OS_SocketDomain_INET, OS_SocketType_STREAM);
OS_SocketBind(Socket, &Addr);
Status = OS_SocketAccept(Socket, &ClientFd, &Addr, 0);

**System observed on:**
Debian 9 64-bit x86

**Additional context**
removing the check for refcount results in expected behavior, OS accept() call should be thread-safe so need not maintain lock through SocketAccept() function.

**Reporter Info**
Chris Knight, NASA Ames

CDKnightNASA:
**Describe the contribution**
Fixes #349, note that this does *not* include unit tests. There don't seem to be any unit tests for the OSAL networking code. We should discuss at the CCB.

**Testing performed**
Tested with the TCP version of SBN, without the fix it fails on the Accept() call, with the fix, SBN nodes connect and traffic flows as expected.

**Expected behavior changes**
OS_SocketAccept() should work to accept new incoming TCP connections.

**System(s) tested on**
64-bit Debian 9 VM.

**Additional context**

**Third party code**
None

Christopher.D.Knight@nasa.gov

### DIFF
--- a/src/os/shared/osapi-sockets.c
+++ b/src/os/shared/osapi-sockets.c
@@ -237,7 +237,7 @@ int32 OS_SocketAccept(uint32 sock_id, uint32 *connsock_id, OS_SockAddr_t *Addr, 
          /* Socket must be of the STREAM variety */
          return_code = OS_ERR_INCORRECT_OBJ_TYPE;
       }
-      else if (record->refcount != 0 || (OS_stream_table[local_id].stream_state & (OS_STREAM_STATE_BOUND | OS_STREAM_STATE_CONNECTED)) != OS_STREAM_STATE_BOUND)
+      else if ((OS_stream_table[local_id].stream_state & (OS_STREAM_STATE_BOUND | OS_STREAM_STATE_CONNECTED)) != OS_STREAM_STATE_BOUND)
       {
          /* Socket must be bound but not connected */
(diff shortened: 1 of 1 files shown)


skliper:
So did this just never work, and never got tested until now?  That's a bit concerning.

astrogeco:
CCB 2020-03-25: Approved

Issue #377, opened 2020-03-23
OSAL Select APIs missing functional test


skliper:
**Is your feature request related to a problem? Please describe.**
OS_SelectMultipe, OS_SelectSingle, OS_SelectFdZero, OS_SelectFdAdd, OS_SelectFdClear, OS_SelectFdIsSet all missing functional tests

**Describe the solution you'd like**
Add tests

**Describe alternatives you've considered**
None

**Additional context**
Certification issue.

**Requester Info**
Jacob Hageman - NASA/GSFC

CDKnightNASA:
I'll start taking a crack at this.

CDKnightNASA:
will wait for #405 to be merged into master and will use new macros, plus need to rebase and remerge the code from #391 and #392

skliper:
This was closed by mistake, Select APIs still need a test to show they function correctly

jphickey:
FYI - there _IS_ now some unit test code which calls "select" .... 

I have to make a change in the select API anyway to fix a different bug (#453), and change I'm making for that issue also requires some updates to these test cases.  This is actually how the bug was noticed -- running this test on RTEMS caused failure.

skliper:
TODO - check if #462 is sufficient to satisfy the unit test standards.

PR #391, opened 2020-03-26
Fix #390, OS_SelectFd... API's check that Set != NULL

### LINKED ISSUE #390: OS_SelectFdZero/OS_SelectFdAdd/OS_SelectFdClear/OS_SelectFdIsSet does not ensure Set is != NULL
**Describe the bug**
Users of these API's will cause a seg fault if they call these functions with a null pointer.

**To Reproduce**
Call OS_SelectFdZero(NULL), etc.

**Expected behavior**
These API's should return an error code when provided a NULL pointer.

**Code snips**
```
int32 OS_SelectFdZero(OS_FdSet *Set)
{  
   memset(Set,0,sizeof(OS_FdSet));
   return OS_SUCCESS;
} /* end OS_SelectFdZero */
```

**System observed on:**
Debian 9

**Additional context**

**Reporter Info**
Christopher.D.Knight@nasa.gov

CDKnightNASA:
**Describe the contribution**
fix to OS_SelectFdZero/Set/Add/Clear/IsSet to check that Set is not NULL

**Testing performed**
Will commit unit tests that check this as a separate pull request for #377 .

**Expected behavior changes**
API calls should return OS_POINTER_ERROR when Set is NULL. (IsSet returns "false".)

**System(s) tested on**
Haven't tested yet, #377 will test.

Christopher.D.Knight@nasa.gov

### DIFF
--- a/src/os/shared/osapi-select.c
+++ b/src/os/shared/osapi-select.c
@@ -59,11 +59,18 @@ int32 OS_SelectSingle(uint32 objid, uint32 *StateFlags, int32 msecs)
    uint32 local_id;
    OS_common_record_t *record;
 
-   return_code = OS_ObjectIdGetById(OS_LOCK_MODE_REFCOUNT, OS_OBJECT_TYPE_OS_STREAM, objid, &local_id, &record);
-   if (return_code == OS_SUCCESS)
+   if(StateFlags == NULL)
    {
-      return_code = OS_SelectSingle_Impl(local_id, StateFlags, msecs);
-      OS_ObjectIdRefcountDecr(record);
+      return_code = OS_INVALID_POINTER;
+   }
+   else
+   {
+      return_code = OS_ObjectIdGetById(OS_LOCK_MODE_REFCOUNT, OS_OBJECT_TYPE_OS_STREAM, objid, &local_id, &record);
+      if (return_code == OS_SUCCESS)
+      {
+         return_code = OS_SelectSingle_Impl(local_id, StateFlags, msecs);
+         OS_ObjectIdRefcountDecr(record);
+      }
    }
 
    return return_code;
--- a/src/os/shared/osapi-select.c
+++ b/src/os/shared/osapi-select.c
@@ -102,8 +102,16 @@ int32 OS_SelectMultiple(OS_FdSet *ReadSet, OS_FdSet *WriteSet, int32 msecs)
  *-----------------------------------------------------------------*/
 int32 OS_SelectFdZero(OS_FdSet *Set)
 {
-   memset(Set,0,sizeof(OS_FdSet));
-   return OS_SUCCESS;
+   int32 return_code = OS_SUCCESS;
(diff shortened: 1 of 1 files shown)


CDKnightNASA:
> A few comments...
> 
> The pattern employed elsewhere for this type of thing is that simple parameter validation is done first and it can use an inline "return" statement. So can't we just add a simple statement:
> 
> ```
> if (Set == NULL)
> {
>     return OS_INVALID_POINTER;
> }
> ```

I thought the CCB frowned upon returns/breaks as they might skip a cleanup action? I actually prefer straight returns...

> Lastly, more of a gripe -- we do realize that this is only checking for one specific bad pointer value, out of a possible >4 billion (32-bit), right? And using CPU cycles to do so? And masking/obfuscating the real error in the process? But if "coding standards" mysteriously says this is better, we do it!

I am also OK if the CCB says "we don't bother to protect against NULL/invalid pointers."

We could get into more exotic options like a "constructor" that adds a header to the struct and then every call can check that the header is there and valid. :D

jphickey:
> > ```
> > if (Set == NULL)
> > {
> >     return OS_INVALID_POINTER;
> > }
> > ```
> 
> 
> I thought the CCB frowned upon returns/breaks as they might skip a cleanup action? I actually prefer straight returns...

Initial/immediate checks of input parameters are OK/preferred to do this way.  The rule of thumb is anything that can be tested/validated in a stateless manner using only constants and local stack variables and no external dependencies, do it immediately upon entry to the function and use an inline `return` if the test fails.

Once the function gets past the point where it starts touching any global state tables, locks, or basically doing anything with an external side effect, it should carry though to the end of the function to make sure anything it did is properly undone, in a consistent manner.

What should be especially avoided is taking/locking some resource and then repeating the cleanup actions each time a validation test is done.  This is a prime opportunity for errors to happen as things evolve (violates DRY principle).

jphickey:
> I am also OK if the CCB says "we don't bother to protect against NULL/invalid pointers."

Yep, I'm not expecting any action on this, which is why I qualified it as a "gripe" and will continue to do so. It is mainly to raise awareness as to the ridiculousness of expending CPU cycles testing for one particular bad value out of the billions of possible bad values, while simultaneously obfuscating the error making the code less debuggable in the process.  (maybe someday the coding standards can evolve...)

Your previous suggestions of putting in an assert-like macro to test for this stuff would be quite useful, because then at least the user has the opportunity to select a more appropriate action if/when these tests fail.

skliper:
Yup, coding standard thing.  Thou shalt check for null pointers passed to APIs... (not in those exact words).

astrogeco:
CCB 2020-04-01 - Discussed, approved as long as it doesn't change API.

CDKnightNASA:
whoops, accidentally closed this one

CDKnightNASA:
actually want to re-do as part of #377  and will simplify to just return rather than nested if()

PR #387, opened 2020-03-25
Fixes #259, updated osapi-filesys.c + ut_osfilesys_diskio_tes.ct to use…

dmknutsen:
**Describe the contribution**
Fixes #259, Updatese osapi-filesys.c + ut_osfilesys_diskio_test.c to use desired return values.

**Testing performed**
1. Ran unit tests

**Expected behavior changes**
The more descriptive return value (OS_ERR_NAME_NOT_FOUND) will now be returned from the following functions (instead of OS_FS_ERROR):
OS_rmfs
OS_mount
OS_unmount
OS_FS_GetPhysDriveName

**System(s) tested on**
Oracle VM VirtualBox
OS: ubuntu-19.10
Versions: cFE 6.7.10.0, OSAL 5.0.8.0, PSP 1.4.7.0,

Contributor Info 
Dan Knutsen
NASA/Goddard

PR #404, opened 2020-04-02
Fix #261, #312, and #362, OSAL build cleanup (multiple issues)

### LINKED ISSUE #261: Improve consistency and reduce duplication in various BSP/PSP implementations
Currently in the CFE/OSAL world we have at least four different platform-specific abstraction methods, all of which basically do the same thing of getting common code to run on a different platforms.  All of these provide three very basic bits of functionality, in different flavors:

- provide a well-known entry point symbol for the platform (main, Init, etc)
- get the system into a proper state for running the app
- Invoke `OS_Application_Startup()` -- or whatever the real app entry point is
- Do something for idle time while the app runs (a shell, or just an idle loop).
- Shut it down when complete 

**Provider 1**:

On the OSAL side this is provided by the "BSP".  It has genppc-vxworks6.4, pc-linux, pc-rtems, sparc-vxworks6.7, sis-rtems, and mcf5235-rtems (the latter 3 are somewhat unmaintained/untested). 

**Provider 2**:

When running Unit tests, an alternative OSAL BSP is used which has extra functionality for running the tests.  The idea is that any platform capable of running applications should also be capable of running Unit tests.

Unfortunately, to avoid changing existing BSP functions, this was "cloned" and therefore this has basically morphed into a second duplicate provider of the entry point/startup/shutdown logic, because UT needs a slightly different setup from a regular app.  There still is, however, a considerable amount of overlap between the "UT" BSP and the regular application BSP.

//Proposal//: These should be made into an extension of the basic OSAL BSP.

**Provider 3**:

The CFE PSP also provides many of the same services.  On the CFE PSP repo, there are "pc-linux", "pc-rtems", and "mcp750-vxworks6.4".

These are basically extended versions of the OSAL BSP, and also provide entry point/shutdown logic, but they also do ADDITIONAL configuration that is specific to CFE before calling the CFE entry point.

//Proposal//: like the unit tests, the CFE PSP should be an extension of the OSAL BSP, not a replacement for it.  The PSP already defines a `CFE_PSP_Main()` entry point for itself, and this can be invoked as a second-stage after OS_Application_Startup to do the additional startup tasks required for CFE.

**Provider 4**:

The JSC-provided unit tests residing in the `osal/src/unit-tests` directory contain their own platform abstractions by way of preprocessor macros and alternate header/source files.  The source/header file choice is based off the `_LINUX_OS_`, `_VXWORKS_OS_`, and `OSP_ARINC653` compile-time macros.  In turn, this defines additional macros for platform-specific routines like sleep / logging / taskDelay, etc.

These tests are already using an entry point provided by the UT-BSP described in (2) above, the macros supplement this.  The problem is that not all platforms are provided.  Most notably, there is no implementation for _RTEMS_OS_, so these tests are currently unbuildable on PC-RTEMS.

//Proposal//: Functions like sleep/delay and other platform-specific bits should be changed from a macro to an API call provided by the UT-BSP.  Logging functions are already provided by the UT-BSP.  Furthermore, all the test cases should use UtAssert calls to make the output consistent with everything else.

### LINKED ISSUE #312: OSAL CMake script should not force/overwrite CMAKE_C_FLAGS
**Describe the bug**
The OSAL build script currently configures several compiler flags by directly setting the `CMAKE_C_FLAGS` variable, and also setting it differently for unit test vs. FSW code.

This was originally done for compatibility with very old CMake versions but this is no longer necessary, as any reasonable CMake version (including v2.8.12 distributed in RHEL/Centos6+) have better commands to deal with target-specific flags (e.g. target_compile_definitions, etc).

Overriding this variable is not ideal as it is expected to retain its value from the parent.

**To Reproduce**
Build OSAL using the current script and ENABLE_UNIT_TESTS=TRUE.  CMAKE_C_FLAGS is forcibly reset twice during the CMakeLists.txt evaluation.

**Expected behavior**
`CMAKE_C_FLAGS` should not be modified by the OSAL build script.  It should preserve whatever value the parent had set (if any) and use the preferred commands (e.g. target_compile_options, etc) to manage the different flags required for UT and normal FSW code.

**System observed on:**
Ubuntu 18.04 LTS, 64 bit

**Reporter Info**

### LINKED ISSUE #362: OSAL coverage test should be part of normal build
Currently the `src/unit-test-coverage` directory is an entirely separate build which needs to be invoked separately by the user.

As per CFE request nasa/cfe#386 it is desired to integrate the OSAL coverage tests as part of the normal unit test builds, so these do not need to be built separately.

The tentatively agreed-upon approach is to build _all_ coverage targets by default, and use the target BSP to run them.

**Additional context**
The side effect is that this may build coverage targets which are not relevant to the deployed platform (e.g. building and running a VxWorks coverage test for a platform that does not use VxWorks).

**Requester Info**

### LINKED ISSUE #363: "duplicate symbol '_main'" on macOS when building tests

This is one of the issues addressed by #352. I am extracting it hereby to be considered separately from the rest of my work of macOS port.
 
I am getting linker errors on macOS because the test targets link in `libosal` library that has a `main()` function and they link `libut_bsp.a` which has a `main()` that launches tests. This causes a conflict with these test target' own main() function as follows:

```
duplicate symbol '_main' in:
    libosal.a(bsp_start.c.o)
    libut_bsp.a(bsp_ut.c.o)
```

Usually the cases like this resolved by extracting the main function from one of the conflicting libraries. In this case it could make sense to extract the `main` function from `libosal` to a separate library such as `libosal-init` or `libosal-main` to avoid any conflicts with test targets. That separate library would then not be linked to the tests and the conflict would go away.

There seems to be no alternative to this behavior on macOS. You either have one `main()` function or there is a conflict.

**Additional context**

This is a more complete fragment of the output:

```
[9/17] Linking C executable timer-test
FAILED: timer-test
: && /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -D_XOPEN_SOURCE=600    -D_LINUX_OS_ -Wl,-search_paths_first -Wl,-headerpad_max_install_names -u _OS_VolumeTable -u _OS_Application_Startup tests/CMakeFiles/timer-test.dir/timer-test/timer-test.c.o  -o timer-test  libut_assert.a libosal.a libut_bsp.a -lpthread -ldl posix-mac-addons/src/mqueue/libmqueue.a posix-mac-addons/src/stubs/librt.a posix-mac-addons/src/pthread/libposix-mac-pthread.a posix-mac-addons/src/semaphore/libposix-mac-semaphore.a posix-mac-addons/src/stubs/libposix-mac-stubs.a posix-mac-addons/src/time/libposix-mac-time.a posix-mac-addons/src/timer/libposix-mac-timer.a && :
duplicate symbol '_main' in:
    libosal.a(bsp_start.c.o)
    libut_bsp.a(bsp_ut.c.o)
ld: 1 duplicate symbol for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
[10/17] Linking C executable bin-sem-test
FAILED: bin-sem-test
: && /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -D_XOPEN_SOURCE=600    -D_LINUX_OS_ -Wl,-search_paths_first -Wl,-headerpad_max_install_names -u _OS_VolumeTable -u _OS_Application_Startup tests/CMakeFiles/bin-sem-test.dir/bin-sem-test/bin-sem-test.c.o  -o bin-sem-test  libut_assert.a libosal.a libut_bsp.a -lpthread -ldl posix-mac-addons/src/mqueue/libmqueue.a posix-mac-addons/src/stubs/librt.a posix-mac-addons/src/pthread/libposix-mac-pthread.a posix-mac-addons/src/semaphore/libposix-mac-semaphore.a posix-mac-addons/src/stubs/libposix-mac-stubs.a posix-mac-addons/src/time/libposix-mac-time.a posix-mac-addons/src/timer/libposix-mac-timer.a && :
duplicate symbol '_main' in:
    libosal.a(bsp_start.c.o)
    libut_bsp.a(bsp_ut.c.o)
ld: 1 duplicate symbol for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
ninja: build stopped: subcommand failed.
```

**Requester Info**

Stanislav Pankevich

jphickey:
**Describe the contribution**

General cleanup and rework of the OSAL build scripts.   Fixes several issues:

Fixes #312 
Fixes #261 
Fixes #362 
Fixes #363 

These are combined into a single pull request as they each depend on the previous fix - they cannot be easily isolated into separate pull requests.  However, each specific issue fix is still an individual commit for review.

Summary of changes:
- Do not clobber (or generally even rely upon) the `CMAKE_C_FLAGS` variable, instead use the preferred approach of using the `target_include_directories` and `target_compile_definitions` functions from CMake to manage the build flags per target.
- Build the implementation components using a separate CMakeLists.txt file per-directory rather than aux_source_directory.  This allow allow BSP-specifc or OS-specific compiler flags to be set, where required.
- Provide sufficient framework for combining the OSAL BSP, UT BSP, and the CFE PSP and eliminating the duplication/overlap between these items.
- Change the OSAL Coverage Test to be built with the standard build whenever ENABLE_UNIT_TESTS is set.

**Testing performed**
- Built CFE for all three supported platforms (native/x86-64 Linux, i686-rtems4.11, ppc-vxworks6.9) and confirm successful build.
- Confirm UT passing on RTEMS 4.11 and native/x86-64 Linux
- Sanity check on CFE (execute, send command) in RTEMS 4.11 and native/x86-64 Linux
- Confirm standalone OSAL builds (without CFE) on basic Posix/PC-Linux combo and passes unit tests.

**Expected behavior changes**
- OSAL coverage testing results shown in standard CFE test build (make test, etc).

**System(s) tested on**
- Ubuntu 18.04 LTS 64-bit (native)
- QEMU for testing RTEMS 4.11 builds

**Additional context**
VxWorks testing limited to building only at this point.  Runtime tests pending on MCP750 hardware availability.


src/os/inc/osapi-os-core.h
```diff
@@ -1505,4 +1505,59 @@ void OS_printf_disable(void);
 void OS_printf_enable(void);
 /**@}*/
 
+
```
skliper, line 1508: Curious on thoughts about these in osapi-os-core.h.  I'd think they'd be candidates for a separate or different header.

jphickey: Yes, these could be split into a separate header file.

Note that OSAL is "caching" the value from the BSP here, so the API really is still an OSAL API, not a BSP API. The BSP sets the cache when it boots and the API just reads it and returns it to the app. I was hoping to avoid having an actual BSP API - the goal should be to hide the BSP details as much as we can, not expose it.

FWIW, we should also consider making the `inc` directory structure more like other modules where we have the public API in `fsw/inc` and the source in `fsw/src`, rather than `src/os/inc` where OSAL has it right now. That would be a follow-on change but this may actually make it easier by removing at least some build system references to `osal/src/os/inc`.

skliper: Thanks, that background helps... historically I've broken includes into smaller "topics" to limit conceptual impact of changes, but I realize OSAL has an established pattern.  It has paid off for me in relation to review panels and operational software support processes.

I think cFS 7.0 is due a major directory restructure... very interested in an in-depth design/review CCB discussion on the topic.

astrogeco:
CCB 20200408 - Concept approved. @acudmore will review by April 13 2020

acudmore:
This is a nice clean up. I like the layering of the OSAL BSP with PSP. Something I have been thinking about for a while but did not quite know how to tackle.
A few Cmake novice questions:
1. In the top level CMakeLists.txt file line you have the add_library( osal_bsp ... ) - Could that be in a CMakeFiles.txt in the directory where the files are? ( just curious of listing files in here vs in the source directory like they are in the src/os/posix) Same applies for the add_library (osal ..) at line 163.
2. is the add_library (<name> OBJECT .. ) just used to create a collection of object files like the old make system did? In other words, not an .so or .a , but just a collection of .o files?
3. Will the files in osal/src/portable get a CMakeLists.txt file when they are no longer included?

jphickey:
Force-push: rebased to current master branch

jphickey:
>     1. In the top level CMakeLists.txt file line you have the add_library( osal_bsp ... ) - Could that be in a CMakeFiles.txt in the directory where the files are? ( just curious of listing files in here vs in the source directory like they are in the src/os/posix) Same applies for the add_library (osal ..) at line 163.

Yes, I suppose it could be... I guess I didn't see the point in splitting those off to a separate file, because they are somewhat fixed and always used.  Where the "impl" stuff has one subdirectory selected out of several choices, the "shared" stuff is always built in the same way from the same set, so splitting it into a separate subdirectory just meant yet another file to look at. 

> 
>     2. is the add_library ( OBJECT .. ) just used to create a collection of object files like the old make system did? In other words, not an .so or .a , but just a collection of .o files?

Yes, pretty much.  an OBJECT library is just a collection of .o files, which are then brought into into the static lib (.a) using TARGET_OBJECTS expression.  I had at one point done everything as .a files but ended up with some order-of-library dependency issues for the final link.  This "object" approach yields fewer different .a files and less issues.

The main thing to note with OBJECT is that it was added to CMake around 2.8.8, which was an issue for RHEL5 which only shipped 2.6.4.  But now that's finally EOL, so RHEL6 is probably the most-ancient tool version used in production, which has 2.8.12, so it should be fine.

>     3. Will the files in osal/src/portable get a CMakeLists.txt file when they are no longer included?

Not exactly, they'll be referenced from the other three implementations in the spots where they are relevant, similar to the way they `#include` them now.  Because one wouldn't necessarily be able to even compile some files on a platform that didn't supply the API needed.

astrogeco:
@acudmore how do you feel about @jphickey's reply?

skliper:
Pushed a fix to resolve coverage test failures introduced by #387, recommending fasttrack

skliper:
Still failing bundle CI, specifically for BUILDTYPE=release, https://travis-ci.com/github/skliper/cFS/builds/160109318 

/home/travis/build/skliper/cFS/osal/src/bsp/pc-linux/src/bsp_console.c: In function ‘OS_BSP_ConsoleOutput_Impl’:
/home/travis/build/skliper/cFS/osal/src/bsp/pc-linux/src/bsp_console.c:68:5: error: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Werror=unused-result]
     write(STDOUT_FILENO, Str, DataLen);
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors

src/bsp/pc-linux/src/bsp_console.c
```diff
@@ -0,0 +1,102 @@
...
+ ****************************************************************************************/
+
+/*----------------------------------------------------------------
+   OS_BSP_ConsoleOutput_Impl
+   See full description in header
+ ------------------------------------------------------------------*/
+void OS_BSP_ConsoleOutput_Impl(const char *Str, uint32 DataLen)
+{
+    /* writes the raw data directly to STDOUT_FILENO (unbuffered) */
+    write(STDOUT_FILENO, Str, DataLen);
```
skliper, line 68: /export/home/jhageman/cFS/cFS-GitHub/osal/src/bsp/pc-linux/src/bsp_console.c:68:5: error: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Werror=unused-result]
     write(STDOUT_FILENO, Str, DataLen);
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors

jphickey: f59b16b puts this write in a loop that checks the return code.  There is still no action taken if it fails (there isn't much to do).

skliper:
Passes now, with https://travis-ci.com/github/skliper/cFS/builds/160232404 which includes nasa/cFE#585, and nasa/psp#149

skliper:
Jumped the gun...

Coverage report no longer includes cFE coverage analysis, and report is misleading for osal since it includes the unit-test-coverage code statistics.  Also actual coverage looks questionable, since not all files included by vxworks look like they are being analyzed.

src/bsp/pc-linux/build_options.cmake
```diff
@@ -0,0 +1,28 @@
...
+)
+
+# C flags that should be used when (re-) compiling code for unit testing.
+# Note: --coverage is just a shortcut for "-ftest-coverage" and "-fprofile-arcs"
+# This also does not work well when cross compiling since paths to the _compile_ dir
+# are baked into the executables, so they will not be there when copied to the target
+# Note - although GCC understands the same flags for compile and link here, this may
+# not be true on all platforms so the compile and link flags are specified separately.
+if (NOT CMAKE_CROSSCOMPILING)
+  set(UT_COVERAGE_COMPILE_FLAGS -pg --coverage)
```
skliper, line 21: This broke cFE coverage reporting.

jphickey: Ack.... I have an update for the CFE side to use UT_COVERAGE_COMPILE_FLAGS.  Will need to be included.

skliper: I was just about to push it into a FASTTRACK for fix-581... want me to hold?

jphickey:
See nasa/cfe#613.....

As the apps now also share this pattern, it might be worth providing a cmake function to build and link a UT.  The alternative might be to just continue setting `UT_C_FLAGS` as equivalent to `UT_COVERAGE_COMPILE_FLAGS`.

src/bsp/pc-linux/build_options.cmake
jphickey: I submitted as a separate pull, not to muddy up 581.  Recommend to fasttrack nasa/cfe#613.

jphickey:
FYI - additional commit added to fixup the previous fixup, I realized it should have a check for DataLen>0 in the loop.  (with a conforming "write()" it should work either way, but this is better).

acudmore:
These changes are OK with me.. I did not mean to hold anything up, my questions were more for my knowledge.

PR #406, opened 2020-04-06
Fix #12, Comment update to correct for microseconds not always rounding up + a…

### LINKED ISSUE #12: Microsecond round up code doesn't round up.
**Describe the bug**
Spawned from #1 

The code comment claims it rounds up to never return zero.  The formula implemented doesn’t actually round up in all cases, since generally when casting a float/double to an int you lose the fractional part (truncation, not rounding).  So the code is not self-consistent.  It’s not a POSIX or OS issue, it’s that the code doesn’t do what it says it does.  The API document doesn’t specify a non-zero guarantee.

https://github.com/nasa/osal/blob/bfa7a33020e9a77221352a99bd2dbee4860ba852/src/os/posix/ostimer.c#L284-L290

Similar misleading comment at:

https://github.com/nasa/osal/blob/bfa7a33020e9a77221352a99bd2dbee4860ba852/src/os/posix/ostimer.c#L231-L232

For what it’s worth, on Linux (our Ubuntu dev system) this code reports 100 ticks per second, and 10000 usec per tick.  But if you pass in high values for ticks per second, it does return zero when it claims to round up (try 2000000 ticks per second).

**To Reproduce**
1. Compile
```
#include <stdio.h>
void main()
{
  float num = 0.7;
  printf("float = %f, cast = %d\n", num, (int)num);
}
```
2. Execute:
```
float = 0.700000, cast = 0
```

**Expected behavior**
Expected code to match comment, round up to not equal zero.  Algorithm doesn't work as claimed in comment.

**System observed on:**
 - cFS development server
 - OS: 4.4.0-146-generic #172-Ubuntu
 - Versions: n/a, sample code shows behavior (although I poked in cFE to cause the same thing)

**Additional context**

**Reporter Info**
Jacob Hageman/NASA-GSFC

dmknutsen:
Describe the contribution
Fixes #12 , Microsecond round up code doesn't round up.
Updated comments in ostimer.c to reflect what the code is actually doing.
Added debug message if microsecs per sec value does not equal 1000000.

Testing performed
Ran unit tests.

System(s) tested on
Oracle VM VirtualBox
OS: ubuntu-19.10
Versions: cFE 6.7.11.0, OSAL 5.0.9.0, PSP 1.4.7.0

Contributor Info
Dan Knutsen
NASA/Goddard

### DIFF
--- a/src/os/shared/osapi-common.c
+++ b/src/os/shared/osapi-common.c
@@ -64,6 +64,7 @@ int32 OS_API_Init(void)
 {
    int32  return_code = OS_SUCCESS;
    uint32 idtype;
+   uint32 microSecPerSec;
 
    if (OS_SharedGlobalVars.Initialized != false)
    {
@@ -159,6 +160,14 @@ int32 OS_API_Init(void)
    {
       OS_DEBUG("Implementation failed to initialize tick time globals\n");
       return_code = OS_ERROR;
+   }
+
+   microSecPerSec = OS_SharedGlobalVars.MicroSecPerTick * OS_SharedGlobalVars.TicksPerSecond;
+	
+   if ( microSecPerSec != 1000000 )
+   {   
+      OS_DEBUG("Warning: Microsecs per sec value of %lu does not equal 1000000 (MicroSecPerTick: %ld   TicksPerSecond: %ld)\n", 
+                (unsigned long) microSecPerSec, (long) OS_SharedGlobalVars.MicroSecPerTick, (long) OS_SharedGlobalVars.TicksPerSecond);
    }
 
    return(return_code);
--- a/src/os/posix/ostimer.c
+++ b/src/os/posix/ostimer.c
@@ -285,9 +285,10 @@ int32 OS_Posix_TimeBaseAPI_Impl_Init(void)
        }
 
        /*
-        * Calculate microseconds per tick - Rounding UP
-        *  - If the ratio is not an integer, this will choose the next higher value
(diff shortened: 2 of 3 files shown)


src/os/shared/osapi-common.c
```diff
@@ -161,6 +162,13 @@ int32 OS_API_Init(void)
       return_code = OS_ERROR;
    }
 
+   microSecPerSec = OS_SharedGlobalVars.MicroSecPerTick * OS_SharedGlobalVars.TicksPerSecond;
+	
+   if ( microSecPerSec != 1000000 )
+   {   
+      OS_DEBUG("Error: Microsecs per sec value of %u does not equal 1000000 \n", microSecPerSec);
```
jphickey, line 169: In this debug statement I'd rather want to know the raw values (MicroSecPerTick, TicksPerSecond) as opposed to the multiplication product.

Also, these are `uint32`, so for printing should use the `%lu` format code and cast it to `(unsigned long)`.

skliper:
Likely will require an update to the coverage testing, can handle with this pull request or open a new issue to address it.  Also sounds more like a "warning" than "error"?

astrogeco:
@jphickey were your requested changes addressed?

jphickey:
Yes, this looks fine now.

skliper:
Fix commit message?

dmknutsen:
Sure thing...what do you want me to change it to?

skliper:
`Fix #12, Add tick to microsec rounding warning, update comments`

PR #423, opened 2020-04-17
Fix #344, Consistent directory entry size limit

skliper:
**Describe the contribution**
Changes FileName in os_dirent_t from OS_MAX_PATH_LEN to OS_MAX_FILE_NAME,
and moves OS_check_name_length into OS_TranslatePath so it is
consistantly applied everywhere. Also fixes the length checks in
OS_check_name_length to account for terminating null.

Unit tests updated to match new directory name limit.
Fix #344

**Testing performed**
Steps taken to test the contribution:
1. CI - https://travis-ci.com/github/skliper/cFS/builds/160789041

**Expected behavior changes**
Individual directory names now limited to OS_MAX_FILE_NAME

**System(s) tested on**
 - Hardware: cFS Dev Server 3
 - OS: Ubuntu 18.04
 - Versions: bundle + this branch

**Additional context**
Note it's based on #421 to avoid conflicts and help with IC merge.

**Third party code**
None

**Contributor Info - All information REQUIRED for consideration of pull request**
Jacob Hageman - NASA/GSFC

PR #421, opened 2020-04-17
Fix #262,  Deprecates OS_FS_* defines that aren't unique to FS

dmknutsen:
**Describe the contribution**
Fix #262 
 Deprecates the following OS_FS* defines from osapi-os-filesys.h:
OS_FS_SUCCESS, OS_FS_ERROR , OS_FS_ERR_INVALID_POINTER, OS_FS_ERR_NO_FREE_FDS , OS_FS_ERR_INVALID_FD, and OS_FS_UNIMPLEMENTED 

Used sed to perform a bulk find/replace on each one to swap with common definitions.

**Testing performed**
Steps taken to test the contribution:
1. Ran unit tests.

**Expected behavior changes**
None

**System(s) tested on**
Oracle VM VirtualBox
OS: ubuntu-19.10
Versions: cFE 6.7.12.0, OSAL 5.0.11.0, PSP 1.4.7.0

**Contributor Info **
Dan Knutsen
NASA/Goddard

PR #427, opened 2020-04-20
Fix #285, Refactor OSAL code selection

### LINKED ISSUE #195: VxWorks OSAL uses potentially unsafe "strcpy" operations
In at several locations the VxWorks OSAL is using `strcat()` and `strcpy()` functions to copy strings between string buffers that are sized using different macros.

At least one of the sizes in play, the `OS_MAX_PATH_LEN` comes from the user-configurable "osconfig.h" file.  Other sizes, such as `OS_FS_PHYS_NAME_LEN` are specified in the local headers and are **not** user-configurable.

In some functions, such as `OS_mkfs` (but not limited to this), a local buffer of size `OS_MAX_PATH_LEN` is copied into a global buffer of size `OS_FS_PHYS_NAME_LEN`.  

However, because the `OS_MAX_PATH_LEN` is configurable via the osconfig.h file, it is not guaranteed that `OS_MAX_PATH_LEN` is less than or equal to `OS_FS_PHYS_NAME_LEN`.

### LINKED ISSUE #214: posix: Use RTLD_DEFAULT instead of NULL in dlsym
In {{{OS_SymbolLookup}}} defined in {{{osloader.c}}} in the POSIX OSAL, [http://pubs.opengroup.org/onlinepubs/9699919799/functions/dlsym.html dlsym] is used to obtain the address of the specified symbol:
{{{
   Function = dlsym((void *)0, SymbolName);
}}}

The behavior of {{{dlsym}}} when called with a null pointer for the first argument is not defined by POSIX. On GNU/Linux, this value corresponds to {{{RTLD_DEFAULT}}}, as found in {{{/usr/include/dlfcn.h}}}.
{{{
# define RTLD_DEFAULT	((void *) 0)
}}}

When called with {{{RTLD_DEFAULT}}} as a first argument, "The identifier lookup happens in the normal global scope; that is, a search for an identifier using handle would find the same definition as a direct use of this identifier in the program code." This appears to be the intended behavior in the OSAL.

However, the value of {{{RTLD_DEFAULT}}} is implementation-defined according to POSIX. The macro, rather than the value it happens to have on GNU/Linux, should be used.

This would address part of #203.

### LINKED ISSUE #285: Replace #include of c files with build system source selection
#include of a c file is not acceptable flight coding style for some organizations and complicates formal verification processes

Source selection should be done via build system

None.

**Additional context**
None.

**Requester Info**
Jacob Hageman - NASA/GSFC

### LINKED ISSUE #432: Simplify config by inferring OSAL_SYSTEM_OSTYPE from OSAL_SYSTEM_BSPTYPE
Currently the build requires that the user specify both OSAL_SYSTEM_BSPTYPE and OSAL_SYSTEM_OSTYPE.  This is redundant because each BSP only works with a specific OS.  

The user should be allowed to specify only OSAL_SYSTEM_BSPTYPE, and the correct OS layer can be inferred automatically.

Continue as is, which increases user config burden and risk of misconfiguration

**Additional context**
see also nasa/cfe#632

**Requester Info**

jphickey:
**Describe the contribution**
Use separate source files and CMake-based source selection based on feature configuration, rather than using the C preprocessor for including/excluding different OSAL function groups.

Refactor all implementation units to provide a separate header file for each functional group/subsystem.  Remove "static" declaration on internal helper functions so they can be invoked from unit test.

Fix #285 (primary)
Also Fix #214 and fix #195 (trivial fixes rolled in as part of refactoring) and fix #432 (separate commit)

**Testing performed**
Build and execute CFE for VxWorks, POSIX and RTEMS.  Sanity check of CFE functions.  Confirm all unit tests passing and coverage test of VxWorks is working (lcov shows 100% for covered modules).

**Expected behavior changes**
No impact to runtime code. Changes build system considerably, however.
- No more user-maintained `osconfig.h` file - now replaced by a cmake configuration file
- Break up low level implementation into small, separate subsystem units, with a separate header file for each one.

**System(s) tested on**
Ubuntu 18.04 LTS 64 bit (build host, native test)
i686-rtems4.11 cross build using QEMU
ppc-vxworks6.9 cross build using MCP750


jphickey:
Force-push to rebase onto latest "master" branch rather than integration-candidate, now showing as one commit...

jphickey:
Force pushed to squash all fixes so far.  Tested on MCP750 and looks good there too.  This is ready for CCB review.

skliper:
20200422 CCB - Covered concepts and patterns, line-by-line and official approval will be off-line.

jphickey:
Question - Just added enhancement #434 for the separate issue, which made me think rather than adding a separate config option for every debug feature, maybe we should also rename `OSAL_CONFIG_DEBUG_PERMISSIVE_MODE` to be simply `OSAL_CONFIG_DEBUG_MODE` so that other debug-related items could be grouped with it rather than making a new option for each debug feature.  I can foresee having other debug niceties be added in the future.

As this PR is already changing these configs, maybe it should preemptively rename this one and extend its scope to be applicable to these other debug features, or is it preferable to have more fine-grained options?

skliper:
I prefer to continue to support custom configuration of the individual options, but have no problem with a general "wrapper" switch to enable them all.

astrogeco:
@skliper based on your email I added the fast track label, do you want to put this in the 20200415 IC? I'll probably merge that to master by Monday April 27

jphickey:
I don't think we can do that due to cross dependencies.  There were already other items in the 2020-04-22 CCB which introduced changes that need to be reconciled/merged with this change set.

I recommend to follow the normal cycle, just do so as expediently as we can.

astrogeco:
@jphickey and @skliper would it be possible to merge just this and https://github.com/nasa/cFE/pull/634 on top of the current integration-candidate (20200415)? There are 20 other tickets on deck for the 20200422 IC and I'd rather get this up and running

jphickey:
My preference would be to get everything ELSE that was approved on 20200422 into the IC, and then I'll rebase this on top of it and re-test everything.

jphickey:
To address Alan's questions:

>Does the loader support symbolic lookup (dlsym) with the static loader? I have an implementation for RTEMS where you can link apps to the OS image and the OSAL uses dlsym to look up the task entry point (does not require the developer/integrator to supply a static symbol table). Using this and a unique path in the startup script (/builtin/sample_app.obj for example), the build can have a combination of built in and statically linked objects.

Yes - but not with dlsym.  One of the driving needs for this feature was a project I worked on that required a small, self-contained virtualized CFS image that could be efficiently live-migrated from one hypervisor to another in real-time(-ish).  To do this, it had to be small, so we went on a space-saving crusade.

The static loader uses a table of "fake" module names, symbol names, and addresses.  When a module load request comes to load something, it first checks this table for a matching module name, and if found, it skips the ModuleLoad_Impl/dlopen and simply returns.  Likewise when a symbol lookup request comes in, it also checks the static table, and the symbol address will be returned from there if found.

So if desired, one can completely compile-out the dynamic loader (dlopen + dlsym) and also compile with visibility=hidden so the entire symbol table doesn't need to be generated or loaded on the target.  

But it is completely transparent to CFS apps - they are still "loaded" and are even assigned a valid module ID.

The limiting factor, of course, is that it only works with symbols that you pre-select.  But in general the only symbol that needs to be there is the entry point for each app.


> If I wanted to add a third party network stack to an OS like LWIP on RTEMS, How would I go about that? Would it involve a new OS configuration option, combined with selecting the new network implementation files in the CMakeLists.txt file?

Yeah, the feature was always intended to support this option but we don't have a good test platform to validate the concept.  

In theory, if you use LwIP with the sockets API, you could just tune the include paths and use the bsd-sockets code as-is.  But more likely you'd want to save the overhead and use the lwip native API, so there would be another alternative to bsd-sockets, like lwip-sockets, in the portable dir.  The CMakeLists would then just have to pick this other file on that platform.  The exact mechanism to do that is open for discussion; it could be another config that the user specifies, or it could be something where the BSP selection hints at the OS layers to build automatically.

Thanks for reviewing!

jphickey:
To clarify that first point - so long as `OSAL_CONFIG_DYNAMIC_LOADER` and `OSAL_CONFIG_STATIC_LOADER` are both enabled, then yes, both dynamic lookups and "real" module loading and symbol lookup via dlsym works, and can be intermixed with static modules that are already loaded and the process is faked - but the same API works transparently for both.

Hope that answers the question.

astrogeco:
@jphickey merged everything else, might need to work with @skliper and @dmknutsen on #423 and #421

jphickey:
FYI - rebased to the 2020-04-15 baseline.  Had to change the base branch back to master too (as it is!).

astrogeco:
Merged in 7a7e6d1f585db719b8914a3697fe76d236b4dee7

PR #436, opened 2020-04-22
Fix #432, Infer OSAL_SYSTEM_OSTYPE from OSAL_SYSTEM_BSPTYPE

### LINKED ISSUE #432: Simplify config by inferring OSAL_SYSTEM_OSTYPE from OSAL_SYSTEM_BSPTYPE

jphickey:
**Describe the contribution**

Simplify user configuration by inferring the `OSAL_SYSTEM_OSTYPE` from the `OSAL_SYSTEM_BSPTYPE` if the former is not set.  This means the user does not have to explicitly set both of these configuration options.

In the event they are _both_ explicitly set (as is done in the current configurations) but refer to an invalid/unexpected combination of OS+BSP, this adds a warning message to the CMake setup (make prep) stage, as this is likely a mis-configuration that deserves attention, but there could be valid reasons for doing this during development.

Fixes #432 

**Testing performed**
Build OSAL for various platform combos (native, x86-linux cross, i686-rtems4.11 cross).  Confirm successful build.  

**Expected behavior changes**
No FSW affected, only build script.  Also fully backward compatible with old configs.

- Builds successfully using the inferred OS when only OSAL_SYSTEM_BSPTYPE is set (not OSAL_SYSTEM_OSTYPE).  
- Generates a warning when OSAL_SYSTEM_BSPTYPE and OSAL_SYSTEM_OSTYPE are both set but are mismatched.  Confirmed this works by intentionally mismatching in a config.

**System(s) tested on**
 - Ubuntu 18.04 LTS 64 bit

**Additional context**
Prerequisite to PR nasa/cfe#634.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,26 +48,6 @@
 cmake_minimum_required(VERSION 2.8.12)
 project(OSAL C)
 
-# OSAL_SYSTEM_OSTYPE and OSAL_SYSTEM_BSPTYPE indicate which of the OS packages
-# to build.  These are required and must be defined. Confirm that this exists
-# and error out now if it does not.
-if (NOT DEFINED OSAL_SYSTEM_OSTYPE OR
-    NOT IS_DIRECTORY "${OSAL_SOURCE_DIR}/src/os/${OSAL_SYSTEM_OSTYPE}")
-  # It is an error if the indicated OSTYPE does not correspond to a subdirectory
-  # If this is not caught here then a more obfuscated error will occur later.
-  message("Error: \"${OSAL_SYSTEM_OSTYPE}\" is not a valid OS type")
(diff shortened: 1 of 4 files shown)


jphickey:
This was originally reviewed in the 20200422 CCB, as commit 2f67435 in PR #427 but was rebased so it can be fast-tracked with nasa/cfe#634 to fix a current issue in the master branch.

PR #442, opened 2020-04-29
Fix #437, deprecate OS task ID and add replacement API for PSP exceptions

### LINKED ISSUE #437: Deprecate OStask_id field from OS_task_prop_t
This field is not necessarily applicable to all OS types and it breaks the abstraction.  OSAL should not be reporting the raw/unabstracted values back to the application.

This field is fundamentally broken in any environment where the underlying OS task ID isn't convertible to a `uint32`.  This includes real systems, such as:

- Cygwin, where `pthread_t` is a compound data type, not an integer (POSIX specifically allows this)
- Systems where the task ID is actually pointer to the TCB (in which case it will be 64 bits on 64 bit systems)

Currently this field serves one purpose, which is to allow the CFE ES exception processing to find the OSAL task ID associated with the exception.   However, this design is being reworked in nasa/cfe#411 where more of the exception processing is done in the CFE PSP.  With this, there should be no need for applications to ever use the `OStask_id` member.

The `OStask_id` member of `OS_task_prop_t` should be marked as deprecated.  As a replacement, an new API can be added which can be invoked by the PSP/BSP to aid in exception processing.  This new API should use an abstract pointer and size, not assume that the task identifier information fits within a 32-bit integer.

**Requester Info**

jphickey:
**Describe the contribution**
The OS_task_prop_t had contained an `OStask_id` field which was used for reverse-mapping the OSAL task ID during exception processing by CFE.  Unfortunately this was a uint32 and not compatible with any OS which used task identifiers that were not convertible to such a value (e.g. `pthread_t`, or 64-bit pointers, etc).

Deprecate the old field and replace with a new API that performs reverse lookup within OSAL using an abstract pointer and size, so its compatible with any OS task representation object.

**Testing performed**
Tested as part of testing exceptions within the scope of PR's nasa/psp#159 and nasa/cfe#653.
Confirm that the newly-added APIs are performing as expected.

**Expected behavior changes**
New API to reverse lookup an OS-provided thread/task identifier back to an OSAL ID.
Any use of existing OStask_id field within the task property structure is now deprecated.

**System(s) tested on**
Ubuntu 20.04 LTS
VxWorks 6.9 on MCP750

**Additional context**
Prerequisite to the PSP/CFE PRs listed above.


jphickey:
Note - this is a "draft" because it is currently pending on #285/#427.  Will rebase and un-draft once #427 is done.

jphickey:
This was closed mistakenly as I meant to rebase but deleted the branch on my fork instead.  Re submitted as #446.

Issue #412, opened 2020-04-10
Implement coverage tests for posix-ng


skliper:
**Is your feature request related to a problem? Please describe.**
Coverage tests not implemented for posix-ng

**Describe the solution you'd like**
Add coverage tests

**Describe alternatives you've considered**
None

**Additional context**
None

**Requester Info**
Jacob Hageman - NASA/GSFC

skliper:
Low priority for now, just adding to track as a future work item

PR #443, opened 2020-05-04
Fix #285, Refactor OSAL to avoid inclusion of C files

### LINKED ISSUE #195: VxWorks OSAL uses potentially unsafe "strcpy" operations

### LINKED ISSUE #214: posix: Use RTLD_DEFAULT instead of NULL in dlsym

### LINKED ISSUE #285: Replace #include of c files with build system source selection

jphickey:
**Describe the contribution**
Use separate source files and CMake-based source selection
based on feature configuration, rather than using the C
preprocessor for including/excluding different OSAL function
groups.

Refactor all implementation units to provide a separate header
file for each functional group/subsystem.  Remove "static"
declaration on internal helper functions so they can be invoked
from unit test.

Fixes #285 
Also Fix #214 and fix #195 (trivial fixes rolled in as part of refactoring)

**Testing performed**
Build and execute CFE for VxWorks, POSIX and RTEMS.  Sanity check of CFE functions.  Confirm all unit tests passing and coverage test of VxWorks is working.  Also confirm build with each inclusion/exclusion of each option (loader, network, shell) and with/without OMIT_DEPRECATED.

**Expected behavior changes**
No impact to runtime code. Changes build system considerably, however.
- No more user-maintained `osconfig.h` file - now replaced by a cmake configuration file
- Break up low level implementation into small, separate subsystem units, with a separate header file for each one.

**System(s) tested on**
Ubuntu 20.04 LTS 64 bit (build host, native test)
i686-rtems4.11 cross build using QEMU
ppc-vxworks6.9 cross build using MCP750


jphickey:
This is a rebase of original PR #427 which was CCB reviewed in a special session on 2020-04-22

jphickey:
> Is osal/src/unit-test-coverage/posix/adaptors/Makefile still used for anything?

Actually the entire POSIX subdirectory in coverage is more or less a placeholder for future use.  We could remove the entire thing with no loss of function, it was only a starting point for someone to actually implement.  

Should I do that?

skliper:
> > Is osal/src/unit-test-coverage/posix/adaptors/Makefile still used for anything?
> 
> Actually the entire POSIX subdirectory in coverage is more or less a placeholder for future use. We could remove the entire thing with no loss of function, it was only a starting point for someone to actually implement.
> 
> Should I do that?

I thought we were using CMake, even for the coverage build?  Just not following why any of the adaptor/Makefile files still exist?  If they are useful or a placeholder, no need to remove.

jphickey:
> I thought we were using CMake, even for the coverage build? Just not following why any of the adaptor/Makefile files still exist? If they are useful or a placeholder, no need to remove.


Correct.  Makefile is not used and not useful in any circumstance.

But on second look, the code in this subdir is also no longer even useful as a template for a future developer because it references files which no longer exist and the README is outdated etc etc.

That's why I'm thinking just remove the entire subdir.  New developer should use VxWorks as the reference point if there is a desire to implement coverage on POSIX.  Existing template for POSIX is probably of negative usefulness as it will point in wrong direction, and no time budget to update it right now.

skliper:
I'd suggest removing whatever Makefile/README's are no longer applicable.  If all of POSIX is out of date feel free to remove.

astrogeco:
@skliper @jphickey can we remove but open an issue to update the template and docs? I figure new developers will probably start on POSIX as opposed to VxWorks?

astrogeco:
@jphickey can we split this up into smaller commits?

skliper:
> @skliper @jphickey can we remove but open an issue to update the template and docs? I figure new developers will probably start on POSIX as opposed to VxWorks?

See #412 as an existing issue to add the POSIX coverage tests.

jphickey:
Force Pushed to split into several commits.  This splits between FSW and coverage test changes.  The bulk of the file-level changes are actually on the unit-test-coverage stuff.

In particular @skliper please have a look at the last commit a9c0205, which is the suggested method to enable permissive mode for CI now.

skliper:
I'm good with a9c0205

PR #448, opened 2020-05-07
Fix #235, add check against OS_MAX_QUEUE_DEPTH

### LINKED ISSUE #235: osconfig.h OS_QUEUE_MAX_DEPTH unused
OS_QUEUE_MAX_DEPTH is defined as 50 in osconfig.h for both pc-rtems and pc-linux, but the limit isn't applied/checked/or even used within OSAL.

Queue depth is accepted as input by OS_QueueCreate with no limiting within OSAL.  Note cFE also has a max pipe depth (CFE_PLATFORM_SB_MAX_PIPE_DEPTH) that is applied by the cFE prior to calling OS_QueueCreate, but it's set to 256.  This limit seems arbitrary at the cFE level.

Linux depth limit by default is 10 on at least CentOS, would be nice if it worked out of the box.

jphickey:
**Describe the contribution**
The `OS_QueueCreate()` function will now sanity check the depth parameter against the configured OS_MAX_QUEUE_DEPTH value.  If it is too large, an error will be returned.  This is a hard limit
and independent of the "permissive" mode.

The OS_MAX_QUEUE_DEPTH should be configured to the largest value that an application may reasonably request.

Fix #235 

**Testing performed**
Confirm CFE framework starts and operates normally (but see nasa/to_lab#36)
Confirm OSAL unit tests pass

**Expected behavior changes**
`OS_QueueCreate()` will return an error code if the depth parameter is larger than the configured `OS_MAX_QUEUE_DEPTH`.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/src/os/inc/osapi-os-core.h
+++ b/src/os/inc/osapi-os-core.h
@@ -507,6 +507,7 @@ int32 OS_TaskGetInfo           (uint32 task_id, OS_task_prop_t *task_prop);
  * @retval #OS_ERR_NAME_TOO_LONG name length including null terminator greater than #OS_MAX_API_NAME
  * @retval #OS_ERR_NO_FREE_IDS if there are already the max queues created
  * @retval #OS_ERR_NAME_TAKEN if the name is already being used on another queue
+ * @retval #OS_QUEUE_INVALID_SIZE if the queue depth exceeds the limit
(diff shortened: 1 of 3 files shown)


skliper:
CCB 20200513 - Approved

PR #461, opened 2020-05-14
Fix #367, 453, 456, 460, Deprecate OS_VolumeTable and other fixes

### LINKED ISSUE #367: OSAL fs defines scrub - missing prefix and unused
FS_BASED, RAM_DISK, EEPROM_DISK, ATA_DISK, NUM_TABLE_ENTRIES all missing OS_ prefix
OS_CHK_ONLY and OS_REPAIR defined, but the input to the expected API is a boolean (bool repair)

Add prefix, remove unused defines

None

**Additional context**
Do after #364 

**Requester Info**
Jacob Hageman - NASA/GSFC

### LINKED ISSUE #453: Portable select() implementation needs to check filehandle "selectable" flag
**Describe the bug**
Not all file handles support the `select()` API -- in particular on RTEMS it only works on sockets.  The File implementation layer contains a boolean flag `selectable` for every file handle to indicate whether the file descriptor can be select'ed upon.

The implementation of this layer needs to confirm/check that the `selectable` flag is true before calling this API.

**To Reproduce**
Run the new `select()` API unit tests on an RTEMS platform, which call this API on regular files.  When it does so, this actually causes an exception and results in kernel panic/abort.    

**Expected behavior**
OSAL should return OS_ERR_NOT_IMPLEMENTED or otherwise avoid calling `select()` on filehandles which the kernel does not handle properly.

**System observed on:**
RTEMS 4.11.3 via QEMU/pc686 BSP.

**Additional context**
Kernel panic result here is rather extreme -- one would have expected that the select() call would simply return error and set an errno as usual.  This probably also signifies an underlying bug in RTEMS.

**Reporter Info**

### LINKED ISSUE #456: Filesystem should differentiate between FS_BASED (known) and unknown filesystem types
The existing `OS_FILESYS_TYPE_DEFAULT` type is used to cover basically any filesystem mapping that doesn't directly fall into the other categories (e.g. NORMAL_DISK, VOLATILE_DISK, etc).  This includes the traditional `FS_BASED` types, which just map to another existing place in the filesystem.

The code should provide a unique value for `FS_BASED` maps, which is a valid mapping type, as opposed to not knowing what mapping is being used.

**Additional context**
Distinction between known FS types and unknown types is required for running without Volume Table in BSP.

**Requester Info**

### LINKED ISSUE #460: Update unit tests to provide consistent directory map
The "file-api-test" and "osal_loader_UT" unit tests rely on a real FS_BASED directory map that is provided by the BSP.  This creates an undesired dependency on the BSP volume table in these test cases, in that they must provide a map for `/cf`.

Test should be BSP-agnostic.   A map for the directory used by the test case can be provided locally as part of this test setup.  A relative system directory can be used (e.g. "./test") to make this platform-agnostic.

**Additional context**
This will allow to run unit tests without a BSP-provided volume table.

**Requester Info**

jphickey:
**Describe the contribution**
Deprecate the OS_VolumeTable structure, which is no longer necessary with the current OSAL.

Remove the OS_VolumeTable definition from all BSPs, and provide a default (empty) one to support linking when OMIT_DEPRECATED is not set.

Note Due to unit test dependencies this PR also has to include fixes for other related issues.

Fix #367 
Fix #453 
Fix #456 
Fix #460

These are all in separate commits and they each build individually but there are some dependencies in making the unit tests work, so they were added to this PR.

**Testing performed**
Build with unit tests, confirm unit test execution and sanity check CFE on supported platforms

**Expected behavior changes**
OSAL no longer defines the types or symbols associated with OS_VolumeTable when OMIT_DEPRECATED is set.  Any external PSPs should replace their respective OS_VolumeTable with calls to the runtime registration API.

**System(s) tested on**
Ubuntu 20.04 (native)
RTEMS 4.11 on i686 via QEMU
Vxworks 6.9 on MCP750

**Additional context**
nasa/psp#161 exists to remove these references from the CFE PSP


astrogeco:
CCB  20200520 - APPROVED

astrogeco:
@acudmore did you take a look at this?

jphickey:
Note -  Lack of this PR is causing the current IC to fail as it is correlated with nasa/psp#163.  We either need to merge this or, if this cannot be merged in this cycle, we need to revert the PSP change.

astrogeco:
> Note - Lack of this PR is causing the current IC to fail as it is correlated with [nasa/PSP#163](https://github.com/nasa/PSP/pull/163). We either need to merge this or, if this cannot be merged in this cycle, we need to revert the PSP change.

I merged the psp one in anticipation that @acudmore would review this one soonish. I think he's working on it today.

PR #462, opened 2020-05-14
Fix #453, Select API and unit test updates

### LINKED ISSUE #453: Portable select() implementation needs to check filehandle "selectable" flag

jphickey:
**Describe the contribution**
On some OS's  (notably RTEMS) one must only call select() socket file handles.  This is indicated in OSAL via the "selectable" flag on each file handle.  The select implementation needs to check this before invoking select().

Also update unit tests to match.

Fix #453

**Testing performed**
Confirm unit tests passing on relevant platforms

**Expected behavior changes**
Calls to select APIs on RTEMS where unsupported will return OS_ERR_OPERATION_NOT_SUPPORTED, rather than crashing.

**System(s) tested on**
Ubuntu 20.04 (native)
RTEMS 4.11 on i686 via QEMU

**Additional context**
This also affects #377 in that it improves the test, but the functional test still could be better in that it currently only checks for file handles being set, not file handles being unset, after select.

### DIFF
--- a/src/os/inc/osapi.h
+++ b/src/os/inc/osapi.h
@@ -77,6 +77,7 @@
 #define OS_ERR_INCORRECT_OBJ_STATE     (-35) /**< @brief Incorrect object state */
 #define OS_ERR_INCORRECT_OBJ_TYPE      (-36) /**< @brief Incorrect object type */
 #define OS_ERR_STREAM_DISCONNECTED     (-37) /**< @brief Stream disconnected */
+#define OS_ERR_OPERATION_NOT_SUPPORTED (-38) /**< @brief Requested operation is not support on the supplied object(s) */
 /**@}*/
 
 /*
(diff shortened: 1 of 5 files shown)


jphickey:
This PR is merged into #461 due to unit test dependencies.  Closing separate PR.

PR #463, opened 2020-05-14
Fix #454, Better error translation on statvfs()

### LINKED ISSUE #454: Improve error code on RTEMS OS_stat() implementation
Depending on the filesystem in use, the RTEMS `statvfs()` call might not be implemented at the filesystem level.  In particular if it is called on the IMFS filesystem type, it returns an error and sets errno to `ENOSYS`.

OSAL translates any error here to `OS_ERROR`, which is what gets returned to the application.

Ultimately this causes the filesystem unit test to fail when this filesystem type is in use.

Preferable to return `OS_ERR_NOT_IMPLEMENTED` in this case.  In particular, unit tests already check for this, and will skip the test cases for this API, avoiding failure.  This makes it a soft error.

**Requester Info**

jphickey:
**Describe the contribution**
Map the ENOSYS errno to OS_ERR_NOT_IMPLEMENTED, which is more appropriate than OS_ERROR.

Fixes #454

**Testing performed**
Build and execute  unit tests on RTEMS 4.11.3

**Expected behavior changes**
Unit test cases which use OS_statfs and run on an RTEMS IMFS volume (default for the root file system) will be skipped and categorized as "NA" due to OS_ERR_NOT_IMPLEMENTED response, rather than a failure.

**System(s) tested on**
RTEMS 4.11 on pc686 via QEMU

### DIFF
--- a/src/os/rtems/src/os-impl-filesys.c
+++ b/src/os/rtems/src/os-impl-filesys.c
@@ -376,17 +376,34 @@ int32 OS_FileSysStatVolume_Impl (uint32 filesys_id, OS_statvfs_t *result)
 {
    OS_filesys_internal_record_t  *local = &OS_filesys_table[filesys_id];
    struct statvfs stat_buf;
+   int32 return_code;
 
    if ( statvfs(local->system_mountpt, &stat_buf) != 0 )
    {
-       return OS_ERROR;
+       /*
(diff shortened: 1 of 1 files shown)


astrogeco:
CCB  20200520 - APPROVED

PR #465, opened 2020-05-14
Fix #457, provide BSP shutdown handler

### LINKED ISSUE #457: Improve ABORT handling in RTEMS unit test
If an unit testing error occurs that causes the test to abort, the implementation just calls the system `abort()` function.  In RTEMS, this causes the entire system/kernel to shutdown.  This leaves no ability to use the shell to check the system state or diagnose what went wrong.

Defer handling of the abort to the BSP.  In RTEMS, when the shell is available, this should just suspend the calling task to stop the tests, allowing the shell to continue being used.

**Additional context**
This used to be the case, but probably an artifact of merging the UT BSP with regular BSP that this got lost in translation, and it only shows up if an abort failure occurs.  Fairly trivial/easy to add it back in though.

**Requester Info**

jphickey:
**Describe the contribution**
This provides a better method of handling test abort, in a BSP-specific manner.

Fixes #457

**Testing performed**
Build and execute unit tests on RTEMS platform, and run a test that has a problem causing an abort.  (e.g. `UtAssert_Abort`).
Confirm that in normal (non-batch) mode the system stays alive and allows use of the shell to check the system state.  When using batch mode the system still shuts down (as expected).

**Expected behavior changes**
No impact to behavior on Linux/VxWorks.
For RTEMS, will not shutdown the kernel if test abort occurs.

**System(s) tested on**
Ubuntu 20.04 (native)
RTEMS 4.11 on pc686 via QEMU

### DIFF
--- a/src/bsp/mcp750-vxworks/src/bsp_start.c
+++ b/src/bsp/mcp750-vxworks/src/bsp_start.c
@@ -24,6 +24,17 @@
 
 #include "mcp750_bsp_internal.h"
 
+/* ---------------------------------------------------------
+    OS_BSP_Shutdown_Impl()
+
+     Helper function to abort the running task
+   --------------------------------------------------------- */
+void OS_BSP_Shutdown_Impl(void)
+{
+    abort();
+}
+
+
 /******************************************************************************
 **  Function:  OS_BSPMain()
 **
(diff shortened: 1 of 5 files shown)


astrogeco:
CCB  20200520 - APPROVED

PR #466, opened 2020-05-15
Fix #459, dynamically create RAM disk devices on RTEMS

### LINKED ISSUE #459: Update RTEMS implementation to dynamically create RAM disk block devices
The current RTEMS BSP predefines the RAM disks by way of the `rtems_ramdisk_configuration` which is set at compile time.  However the OSAL model is to create these block devices at runtime based on memory segments that are allocated externally.

The workaround up to this point was to preallocate the ram disks, and then attempt to correlate the address in the OS_mkfs/initfs request to one of the preallocated blocks.  However this is not possible when the compile-time config and ramdisk allocation are done by separate libraries (BSP and CFE PSP, respectively).

RTEMS can support dynamically-created RAM disks using `ramdisk_allocate()` which better aligns with the way things are supposed to work.

**Additional context**
This is necessary as part of decoupling the volume/filesystem table from the OSAL BSP.

**Requester Info**

jphickey:
**Describe the contribution**
Rather than relying on the BSP to preallocate, the ram disk block devices can be created based on request.  This correlates with the way RAM disks are implemented on VxWorks and is
cleaner and more flexible by making it more independent of the BSP.

This allows all the unit tests to work on RTEMS without the needing the BSP to preallocate a ramdisk.

Fixes #459 

**Testing performed**
Build and run all unit tests, and sanity check CFE operation

**Expected behavior changes**
Unit tests work on RTEMS without BSP preallocating ramdisks

**System(s) tested on**
RTEMS 4.11 on pc686 via QEMU

### DIFF
--- a/src/bsp/pc-rtems/src/bsp_start.c
+++ b/src/bsp/pc-rtems/src/bsp_start.c
@@ -48,29 +48,6 @@ extern rtems_status_code rtems_ide_part_table_initialize(const char *);
 extern int rtems_rtl_shell_command (int argc, char* argv[]);
 
 /*
- * The RAM Disk configuration.
- */
-rtems_ramdisk_config rtems_ramdisk_configuration[RTEMS_NUMBER_OF_RAMDISKS];
-
-/*
- * The number of RAM Disk configurations.
- */
-size_t rtems_ramdisk_configuration_size = RTEMS_NUMBER_OF_RAMDISKS;
-
-/*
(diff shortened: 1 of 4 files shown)


astrogeco:
CCB  20200520 - APPROVED

acudmore:
This is great. I was going to suggest something similar with my GTOSat RTEMS 5 merge, but will rebase my changes on this instead.

PR #468, opened 2020-05-16
Fix #450, add external source directory for OS/BSP

### LINKED ISSUE #450: Add alternative source directory option for OSAL implementations
The OSAL build script always looks for BSP implementations in:
    `${OSAL_SOURCE_DIR}/src/bsp/${OSAL_SYSTEM_BSPTYPE}`
Likewise it always looks for OS implementations in:
    `${OSAL_SOURCE_DIR}/src/os/${OSAL_SYSTEM_OSTYPE}`

This presents a challenge to for a user with a custom OS/BSP implementation not within the mainline OSAL source tree, but wishes to use the github repo directly.

It should be possible to provide an external directory that the OSAL build can use.

Create an OSAL fork or "git subtree" to assemble a full CFE/CFS repo, which allows one to add custom packages or patches they need.

**Additional context**
The internal API between the BSP/OS implementation components is not stabilized like the public API is, and can change any time.  

If directly using a "master" branch from github in conjunction with a locally-controlled implementation module, users may experience frequent breakage, as the component interface can change any time rendering it incompatible with their local version.  

In contrast, although the fork/subtree approach does have a similar issue, it doesn't happen "automatically" - it requires a specific user action to pull and merge the new "master" so it creates a buffer that avoids unsolicited breakage.

This feature should come with a big disclaimer that says its for debug/development only, not to use it in production.

**Requester Info**

jphickey:
**Describe the contribution**

If the "OSAL_EXT_SOURCE_DIR" cache variable is set, this location will be checked first for a BSP/OS implementation layer.  This can point to an out-of-tree implementation layer if necessary.

Fixes #450

**Testing performed**
Configure a test project to use a BSP provided in the external source dir, and confirm that it builds correctly and sanity check result (unit tests, CFE runs normally).

**Expected behavior changes**
No impact to behavior unless user requires a BSP/OS implementation which is not part of the included set.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
It is discouraged from actually doing this as there is no attempt at API stability at the low level implementation layer.

The actual location of the implementation layer being used is now also displayed during "prep" so that one can confirm its the correct/expected one.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,15 @@
 cmake_minimum_required(VERSION 2.8.12)
 project(OSAL C)
 
+# The "OSAL_EXT_SOURCE_DIR" cache variable may be set to a path
+# on the host containing extra OS/BSP implementations which are not
+# part of the open source release.
+# CAUTION: The API between the OSAL and the low level implementation and/or BSP
+# is not stabilized, and may change with every OSAL release.  No attempt is made
+# to provide backward compatibility with to external sources. 
+set(OSAL_EXT_SOURCE_DIR "$ENV{OSAL_EXT_SOURCE_DIR}" 
+    CACHE PATH "External source directory to check for additional OS/BSP implementations")
+    
 # Read the default compile-time configuration, and update with
 # any mission/project specific options in the OSAL_CONFIGURATION_FILE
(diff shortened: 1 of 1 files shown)


astrogeco:
CCB  20200520 - APPROVED

PR #472, opened 2020-05-19
Fix #470, Binary sem task delete issues

### LINKED ISSUE #470: Binary Semaphore locked after thread cancellation
**Describe the bug**
On POSIX systems using the Binary Semaphore API, there is a risk that threads can become deadlocked after deleting a task that was executing an `OS_BinSemTake()` API call.

**To Reproduce**
- Create two tasks and one binary semaphore.
- Task A calls `OS_BinSemGive()` at some event/interval.
- Task B waits on the binary sem using `OS_BinSemTake()`

Then delete task B while it was pending in `OS_BinSemTake()`.

The semaphore resource is now inoperable, because the condition mutex was "owned" by task B at the time it was deleted, and is never released, thereby preventing any other task from using the mutex.

**Expected behavior**
The semaphore should continue to be usable by other tasks after deleting task B.

**System observed on:**
Ubuntu 20.04

**Additional context**
This was originally reported/observed in nasa/cfe#701, during shutdown where timers were used.

**Reporter Info**
(After diagnosis/investigation of issue reported by @excaliburtb in above ticket)

jphickey:
**Describe the contribution**
Corrects issue when a task waiting on a binary semaphore is deleted, it left the mutex in a locked state preventing other tasks from using the mutex.

This has 3 parts:
1. A functional test enhancement to bin-sem-test which replicates the specific conditions for the observed bug to occur.  It deletes the task calling `OS_BinSemTake()` and then attempts to use the semaphore after this.  (without the rest of this change, it fails).
2. Employ a pthread "cleanup handler" to handle the situation where a task is canceled during the `pthread_cond_wait()` call.  This ensures that the mutex is unlocked as part of the cleanup, so other tasks may continue using the semaphore.  This is the main fix as it directly addresses the root cause of the issue, which is that the mutex was left in a locked state.
3. Change all initial mutex locking to be a finite "timed" wait rather than an infinite wait.  In all cases, the condition variable is only held for brief periods of time and should be readily available; tasks should almost never block on this initial lock.  If a task blocks for a long time, this considers the mutex "broken" and aborts, thereby avoiding deadlock.  This is a "contingency" fix in that if an exception or signal or other unknown/unhandled async event occurs that leaves the mutex permanently locked, it at least does not deadlock the system and allows it to be restarted.

Fixes #470 
Fixes nasa/cfe#701

**Testing performed**
Build and run all unit tests.

Confirm that including only the unit test change (item 1 above) reliably reproduces the failure.  In this mode, bin-sem-test will get stuck when attempting to give the semaphore after task deletion.

Confirm that the bin sem changes (items 2 and 3 above) correct the issue.  Note each one taken individually will avoid deadlock in a different way.  Item 3 (timed mutex) alone will cause the subsequent calls to return `OS_SEM_FAILURE` rather than deadlock, so shutdown will continue and the test exits with a failed status rather than pending forever.

Sanity check CFE operation and CTRL+C handling - all works OK.

**Expected behavior changes**
Binary semaphores after task deletion continue to work as expected and are usable by other tasks.

**System(s) tested on**
Ubuntu 20.04

**Additional context**

**Third party code**
If included, identify any third party code and provide text file of license

### DIFF
--- a/src/os/posix/src/os-impl-binsem.c
+++ b/src/os/posix/src/os-impl-binsem.c
@@ -28,8 +28,37 @@
 #include "os-shared-binsem.h"
 #include "os-impl-binsem.h"
 
+/*
+ * This controls the maximum time the that the calling thread will wait to
+ * acquire the condition mutex before returning an error.
+ *
+ * Under normal conditions, this lock is held by giving/taking threads very
+ * briefly, so the lock should be available with minimal delay.  However,
+ * if the "taking" thread is canceled or exits abnormally without releasing the
+ * lock, it means any other task accessing the sem can get blocked indefinitely.
+ *
+ * There should be no reason for a user to configure this, as it should
+ * not be relevant in a normally operating system.  This only prevents a
+ * deadlock condition in off-nominal circumstances.
+ */
+static const struct timespec OS_POSIX_BINSEM_MAX_WAIT =
+{
+        .tv_sec = 2,
+        .tv_nsec = 0
+};
+
+
 /* Tables where the OS object information is stored */
 OS_impl_binsem_internal_record_t    OS_impl_bin_sem_table       [OS_MAX_BIN_SEMAPHORES];
+
+/*---------------------------------------------------------------------------------------
+ * Helper function for releasing the mutex in case the thread
+ * executing pthread_condwait() is canceled.
+ ----------------------------------------------------------------------------------------*/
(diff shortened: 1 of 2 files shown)


jphickey:
@excaliburtb this should resolve nasa/cfe#701

astrogeco:
CCB  20200520 - APPROVED

PR #475, opened 2020-05-19
Fix #474, Add timecb global mutex

### LINKED ISSUE #474: Global lock for "timecb" objects missing
**Describe the bug**
The internal table mutex for the timer callback (timecb) object type is missing/not implemented.   This leave a potential race condition as timer callbacks are registered.

**To Reproduce**
Not observed in real execution at this time - but theoretically possible if two timers are registered by two tasks at the exact same time.

**Expected behavior**
Should be protected.

**System observed on:**
Ubuntu 20.04

**Additional context**
The global lock table is simply missing an entry for this object type.

**Reporter Info**

jphickey:
**Describe the contribution**
Adds the mutex to protect the timer callback (timecb) resource table.

Fixes #474 

**Testing performed**
Build and confirm all unit tests pass, CFE runs normally.

**Expected behavior changes**
Timer callback (timcb) resource types are mutex protected.

**System(s) tested on**
Ubuntu 20.04 
RTEMS 4.11 on pc686 via QEMU

### DIFF
--- a/src/os/posix/src/os-impl-idmap.c
+++ b/src/os/posix/src/os-impl-idmap.c
@@ -42,6 +42,7 @@ static POSIX_GlobalLock_t OS_count_sem_table_mut;
 static POSIX_GlobalLock_t OS_stream_table_mut;
 static POSIX_GlobalLock_t OS_dir_table_mut;
 static POSIX_GlobalLock_t OS_timebase_table_mut;
+static POSIX_GlobalLock_t OS_timecb_table_mut;
 static POSIX_GlobalLock_t OS_module_table_mut;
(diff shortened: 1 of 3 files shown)


skliper:
Does this mean everywhere it was requesting the lock/unlock it was really just a no-op and returning error (since mut == null)?

jphickey:
Yes, but only for the "TimeCB" object table -- which typically is only used once at startup (when still single threaded) and maybe once in SCH.  So low impact as far as usage goes.  Surprised that this was not noticed before....

astrogeco:
CCB  20200520 - APPROVED

PR #478, opened 2020-05-19
Fix #476, add global lock/unlock wrapper

### LINKED ISSUE #476: Add shared layer API for handling global mutex
Currently the shared layer calls `OS_Lock_Global_Impl()` and `OS_Unlock_Global_Impl()` directly when locking/unlocking the internal tables within OSAL.  

This does not offer a place to:
- check return code of operation
- track ownership of resource in case of not being freed.

Should implement `OS_Lock_Global()` and `OS_Unlock_Global()` wrappers at the shared API level, which can provide a common place to check status, handle failure, and track resource usage.

**Additional context**
In #474, the global table lock was correctly returning an error but nothing ever checked the return code.  Rather than adding a specific return code test everywhere this is used, simpler and better to add a wrapper.

**Requester Info**

jphickey:
**Describe the contribution**
Add a wrapper at the shared layer to provide a common location to check the status of global lock/unlock ops.

All calls to OS_Lock_Global_Impl and OS_Unlock_Global_Impl from the shared modules are replaced with calls to this wrapper.

Notably - this detects and reports the problem described in #474, in case something similar happens in the future.  

Fixes #476

**Testing performed**
Build and execute all unit tests, and sanity check CFE operations.

**Expected behavior changes**
Now detects and reports errors in global mutex lock/unlock ops.

**System(s) tested on**
Ubuntu 20.04 (native)
RTEMS 4.11 on pc686 via QEMU

**Additional context**
If this patch is applied in isolation without any other changes, one will see debug trace generated as it is detecting the problem previously reported in #474:

```
OS_Lock_Global():634:ERROR: unable to lock global 9, error=-1
OS_Unlock_Global():677:ERROR: global 9 released by task 0x10005 when owned by task 0x0
OS_Unlock_Global():693:ERROR: unable to lock global 9, error=-1
```

These errors go away as soon as the PR #475 is also applied

### DIFF
--- a/src/os/posix/src/os-impl-idmap.c
+++ b/src/os/posix/src/os-impl-idmap.c
@@ -80,14 +80,7 @@ int32 OS_Lock_Global_Impl(uint32 idtype)
    POSIX_GlobalLock_t *mut;
    sigset_t previous;
 
-   if (idtype < MUTEX_TABLE_SIZE)
-   {
-      mut = MUTEX_TABLE[idtype];
-   }
-   else
-   {
-      mut = NULL;
-   }
+   mut = MUTEX_TABLE[idtype];
 
    if (mut == NULL)
    {
--- a/src/os/shared/inc/os-shared-idmap.h
+++ b/src/os/shared/inc/os-shared-idmap.h
@@ -102,7 +102,16 @@ int32 OS_ObjectIdInit                (void);
 
 
 /*----------------------------------------------------------------
-   Function: OS_Lock_Global_Impl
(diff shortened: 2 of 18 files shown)


jphickey:
Note - this still needs a coverage test update, but the general idea can be discussed.

astrogeco:
CCB  20200520 - APPROVED

astrogeco:
@jphickey can you look and resolve these conflicts?

jphickey:
Manually merged with conflicts resolved in 5c8a01c

PR #480, opened 2020-05-21
Fix #479, add name association to stub arguments

### LINKED ISSUE #479: Ut Assert argument name association
UT assert passes "static" function arguments to hook functions as an array of opaque pointers, based on the order of calls to the `UT_Stub_RegisterContext()` API inside the stub.  This is basically equivalent to argc/argv on a command line.

This requires test cases to "know" which position the argument they need to access is in, and basically hard-code that number.  This could become more problematic if hook functions are used more frequently.

Offer a variant of `UT_Stub_RegisterContext()` that also stores the argument name, and offer a "getter" that a hook function can use in order to get by that name.

Registering and retrieving values by name will provide greater stability as the APIs and test cases evolve over time.

The downside is that its slower and uses more memory to look up based on name, but that doesn't matter much for UT.

Leave as is using number/sequence based args.  It works but concerned it's too easy to get out of sync.

**Additional context**
Could also use some macros to make both the register (stub-side) and get (hook-side) operations a little more user-friendly.

**Requester Info**

jphickey:
**Describe the contribution**
Add the capability to store parameter names in addition to pointers as part of the stub argument context data.

Macroize the UT_Stub_RegisterContext function utstubs.h to also pass the parameter name in addition to the pointer.

Add hook-side accessor functions/macros to make it easier and more reliable to get this information from the context data.

Also other minor items:
- Increase max "fixed" args from 4 to 8
- Add convenience macros for variable argument stubs

Fixes #479 

**Testing performed**
Build and execute unit tests, confirm all passing (baseline).
Update EVS stub and SAMPLE_APP test cases to use new functions and confirm operation.

**Expected behavior changes**
New features only, does not change existing behavior.
UT Hook functions now have the capability to get argument values by name, which is more future proof than assuming a numeric index.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/ut_assert/inc/utstubs.h
+++ b/ut_assert/inc/utstubs.h
@@ -53,8 +53,29 @@ typedef cpuaddr UT_EntryKey_t;
  * Maximum size of a callback hook context list
  *
  * This is the maximum number of function arguments that can be passed to a hook
+ * Note that OS_TaskCreate() has (possibly) the highest parameter count in OSAL with 7 parameters
  */
-#define UT_STUBCONTEXT_MAXSIZE      4
+#define UT_STUBCONTEXT_MAXSIZE      8
+
+/**
+ * Identifies the type of value stored in the ArgPtr field of a UT_StubContext_t object
+ */
+typedef enum
+{
+    UT_STUBCONTEXT_ARG_TYPE_UNSPECIFIED = 0,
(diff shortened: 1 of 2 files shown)


astrogeco:
CCB 2020-05-27 APPROVED

PR #495, opened 2020-06-09
Fix #493, race condition in idmap api test

### LINKED ISSUE #493: Race condition in new idmap tests
**Describe the bug**
The `idmap-api-test` routine occasionally fails, because it creates one of its semaphores in a sub task, but does not actually wait for that sub task to run.

**To Reproduce**
Run idmap test, in particular a single core machine seems to reproduce it more reliably.

**Expected behavior**
Need to add a wait loop of some type in main task to make sure child task has executed.

**Code snips**
Task started here:
https://github.com/nasa/osal/blob/360438765d7eb83a9e7e15571cda89786f0d8330/src/tests/idmap-api-test/idmap-api-test.c#L106

Might want to employ a loop similar to what is done here to make sure the child task has executed:
https://github.com/nasa/osal/blob/360438765d7eb83a9e7e15571cda89786f0d8330/src/tests/osal-core-test/osal-core-test.c#L154-L159

**System observed on:**
CentOS 6.9 VM (single core)

**Reporter Info**

jphickey:
**Describe the contribution**
Add a delay loop to ensure that the child task has executed before continuing with the rest of the test.

This also uses a separate "setup" function to create the resources for the test and adds asserts that the setup worked.

Fixes #493

**Testing performed**
Execute unit tests and confirm passing.

**Expected behavior changes**
No more race condition in idmap-api-test.

**System(s) tested on**
Ubuntu 20.04 (native)
Centos 6.9 VM (single core)

**Additional context**
May want to fast track as this causes occasional CI failures in current master branch.

### DIFF
--- a/src/tests/idmap-api-test/idmap-api-test.c
+++ b/src/tests/idmap-api-test/idmap-api-test.c
@@ -26,6 +26,17 @@
 #include "utbsp.h"
 
 
+uint32 task_id;
+uint32 queue_id;
+uint32 count_sem_id;
+uint32 bin_sem_id;
+uint32 mutex_id1;
+uint32 mutex_id2;
+uint32 mutex_id3;
+uint32 time_base_id;
+
+#define UT_EXIT_LOOP_MAX    100
+
 /* *************************************** MAIN ************************************** */
 
 typedef struct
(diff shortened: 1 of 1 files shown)


astrogeco:
@skliper any objections with Fast Tracking this?

skliper:
> @skliper any objections with Fast Tracking this?

Approved.

PR #500, opened 2020-06-12
Fix #499, store all arguments in generic-linux BSP

### LINKED ISSUE #499: Linux BSP shouldn't trim first argument
**Describe the bug**
As part of command line processing the logic for Linux BSP is to prune the first argument which is the command name, to be consistent with other (e.g. RTEMS) platforms that do not pass the command name to begin with.

However the CFE PSP passes the argument list to the C library `getopt_long()` and this expects the first argument to be the command name, so this breaks.

**To Reproduce**
Pass arguments (e.g. `-I' to set spacecraft ID value, etc) to start the CFE core on linux.  Observe that the default value is still used, argument isn't processed.

**Expected behavior**
Command line arguments should be processed and used.

Just store/pass the full set of arguments, no real need to abstract, as anything doing argument processing is almost certainly platform-specific (e.g. CFE PSP) so it will know what to expect here.

**System observed on:**
Ubuntu 20.04

**Additional context**
First bug found by newly-minted PSP functional test.  I attempted to use `-I` to force a test failure on the spacecraft ID value and it didn't work.

**Reporter Info**

jphickey:
**Describe the contribution**

Keep the entire argc/argv from the shell.  Do not prune the command name as getopt expects this to be there.

Fixes #499 

**Testing performed**
Execute unit tests, confirm all passing.
Execute new PSP tests, confirm command line options are being recognized.

**Expected behavior changes**
Command line options in Linux are recognized, no longer ignored/dropped.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/src/bsp/generic-linux/src/bsp_start.c
+++ b/src/bsp/generic-linux/src/bsp_start.c
@@ -140,14 +140,12 @@ int main(int argc, char *argv[])
      * Note that the first argument (0) is the command name.  The
      * first "real" argument is at position 1.
      *
-     * The first arg is ignored to be more consistent with other platforms
-     * where this is not passed in.
+     * However this still needs to pass it through as the appliction
+     * might still want to use library "getopt" and this expects the
+     * first parameter to be this way.
      */
-    if (argc > 1)
-    {
-        OS_BSP_Global.ArgC = argc - 1;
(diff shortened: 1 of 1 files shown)


skliper:
Ping @SweeWarman

astrogeco:
CCB 2020-06-17: APPROVED

PR #503, opened 2020-06-16
Fix #502, Add extra assert macros

### LINKED ISSUE #502: Add int32 UT assert macro
A very common use case in unit testing is to assert on an `int32` value, as this is a common return value type for error codes and status values.

UT Assert provides a generic `UtAssert_True()` macro which can certainly test for integer equality, but it is still very repetitive and test cases may not include all the relevant information.  For instance, the framework should encourage/require that the actual values are printed in addition to the pass/fail status, and the generic macro being totally free-form does not enforce this at all.

Add a macro `UtAssert_INT32_EQ()` to check for equality of two values as `int32` types.

Other types could be added too, but `int32` is by far the most commonly needed as it is the return type of many API calls across OSAL and CFE.

Continue using the generic `UtAssert_True()` macro.

**Additional context**
Many test programs already use their own macros for testing return values.  They've already evolved to be similar but different (e.g. CFE SB has `ASSERT_EQ`, and OSAL coverage testing has `OSAPI_TEST_FUNCTION_RC()`).  

It would be worthwhile to put more of these in UT assert itself so these don't need to continue being duplicated.

**Requester Info**

jphickey:
**Describe the contribution**
Add the following macros for UT assert:

`UtAssert_INT32_EQ` - check equality as 32 bit signed int
`UtAssert_UINT32_EQ` - check equality as 32 bit unsigned int
`UtAssert_NOT_NULL` - check pointer not null
`UtAssert_NULL` - check pointer is null
`UtAssert_NONZERO` - check integer is nonzero
`UtAssert_ZERO` - check integer is zero
`UtAssert_STUB_COUNT` - check stub count

These are all just wrappers around `UtAssert_True` for commonly-used asserts.  The description message is auto generated so it is consistent.

Fixes #502 

**Testing performed**
Build and run all existing unit tests.
Also test with new PSP coverage tests that use the new macros and confirm they work as described.

**Expected behavior changes**
No impact to current behavior.  These are just new macros for new test case, to improve ease of use and consistency for future testing.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/ut_assert/inc/utassert.h
+++ b/ut_assert/inc/utassert.h
@@ -134,6 +134,84 @@ typedef struct
 #define     UtAssert_Type(Type,Expression,...)      \
                 UtAssertEx(Expression, UTASSERT_CASETYPE_##Type, __FILE__, __LINE__, __VA_ARGS__)
 
+/**
+ * \brief Compare two values for equality with an auto-generated description message
+ * Values will be compared in an "int32" type context.
+ */
+#define UtAssert_INT32_EQ(actual,expect)  do    \
+{                                               \
+    int32 rcexp = (int32)(expect);              \
+    int32 rcact = (int32)(actual);              \
+    UtAssert_True(rcact == rcexp, "%s (%ld) == %s (%ld)",   \
+        #actual, (long)rcact,                               \
+        #expect, (long)rcexp);                              \
+} while(0)
+
+/**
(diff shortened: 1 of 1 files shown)


astrogeco:
CCB 2020-06-17: APPROVED and FastTracked

PR #492, opened 2020-06-08
Fix #491, increase UT assert buffer sizes

jphickey:
**Describe the contribution**
The buffer for context info increases from 128 to 256 bytes and the total report buffer to 320 bytes.

Fixes #491 

**Testing performed**
Build and Execute all unit tests.

**Expected behavior changes**
Confirmed that long report lines that had been truncated are no longer truncated.

**System(s) tested on**
Ubuntu 20.04

**Contributor Info - All information REQUIRED for consideration of pull request**
Joseph Hickey, Vantage Systems, Inc.

PR #504, opened 2020-06-16
Fix #501, add PIC library for ut assert

### LINKED ISSUE #501: Offer secondary target for UT assert library that can be dynamically loaded
The current UT assert is designed to be statically linked as an OSAL application.

To support a CFE functional test environment that uses the same UT assert reporting methods, the library should offer a target that can be linked as loadable module.

The differences are:

- Compile as position independent code (PIC) so it can be linked into a dynamic module.
- Separate out the OS_Application_Startup and OS_Application_Run - as this will be loaded into a larger app, not a standalone app, and these symbols are already defined.
- Consolidate global variables into a common structure and make early init/reset of these variables into a separate routine from OS_Application_Startup.

**Additional context**
This dynamically-loadable variant can be a separate target, so it shouldn't change anything about the existing apps/tests that link with the existing static "ut_assert" library.

**Requester Info**

jphickey:
**Describe the contribution**
Add a position independent code (PIC) variant of the ut_assert library, which can be dynamically loaded into other applications rather than running as a standalone OSAL application.  This enables loading UT assert as a CFE library.

This required moving some symbols/functions around.  Mainly, the PIC library does _not_ contain the `utbsp.c` file which has the bindings to the OSAL BSP when running as a standalone app, but everything else should be the same.

Fixes #501 

**Testing performed**
Build and execute all unit tests.
Also test linking new PIC library into a CFE loadable app and running tests.

**Expected behavior changes**
No impact to current unit testing which runs UT assert as a standalone app.  

**System(s) tested on**
Ubuntu 20.04

**Additional context**
Some functions were moved to support the alternative linking structure but no API/functionality changes.  The moved functions include the buffer length extension that were originally part of PR #492.

**Third party code**
If included, identify any third party code and provide text file of license

### DIFF
--- a/ut_assert/CMakeLists.txt
+++ b/ut_assert/CMakeLists.txt
@@ -42,4 +42,25 @@ target_compile_definitions(ut_assert PUBLIC
 target_link_libraries(ut_assert osal_bsp)
 
 
+# The "pic" variant of ut_assert is compiled as an
+# object library to be included in another object,
+# such as a loadable test app for CFE.
+# It is compiled as position independent code (PIC)
+# to support dynamic loading.
+add_library(ut_assert_pic OBJECT EXCLUDE_FROM_ALL
+    src/utassert.c
+    src/utlist.c
+    src/utstubs.c
+    src/uttest.c
+    src/uttools.c
+)
+
+set_target_properties(ut_assert_pic PROPERTIES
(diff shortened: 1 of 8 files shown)


astrogeco:
CCB 2020-06-17: APPROVED

PR #508, opened 2020-06-16
Fix #506, enforce nonzero stack size

### LINKED ISSUE #506: Idmap API test needs to use real task stack size and priority
**Describe the bug**
Although Linux allows the stack size to be zero (it uses a nonzero default) other OS's do not have this fallback.  The `idmap-api-test` attempts to create a task with zero stack and this fails on RTEMS.

**To Reproduce**
Run 'idmap-api-test` on RTEMS.  `OS_TaskCreate()` fails (reports as TSF) but then other failures occur related to this.

**Expected behavior**
Should pass.

**System observed on:**
RTEMS 4.11.3 running in QEMU.

**Additional context**
Should consider forcibly making every OS fail (i.e. enforce in shared layer) if passed a zero stack.

**Reporter Info**

jphickey:
**Describe the contribution**
Resolve inconsistency in how the stack size is treated across different OS implemntations.  POSIX would enforce a minimum, where RTEMS would not.

With this change the user requested size is passed through to the underlying OS exactly as is, no enforced minimum.

An additional sanity check is added at the shared layer to ensure that the stack size is never passed as 0.

Fixes #506 

**Testing performed**
Build and execute CFE and sanity test
Build and execute all unit tests.
Also tested creating a task with zero stack size and confirm error is generated on Linux.

**Expected behavior changes**
The `idmap-api-test` passes on RTEMS.
Attempting to create a task with zero stack now will result in error.

**System(s) tested on**
Ubuntu 20.04
RTEMS 4.11.3 via QEMU/pc686

### DIFF
--- a/src/os/posix/src/os-impl-tasks.c
+++ b/src/os/posix/src/os-impl-tasks.c
@@ -31,13 +31,6 @@
 #include "os-shared-task.h"
 #include "os-shared-idmap.h"
 
-/*
- * Defines
- */
-#ifndef PTHREAD_STACK_MIN
-#define PTHREAD_STACK_MIN   (8*1024)
-#endif
-
 /* Tables where the OS object information is stored */
 OS_impl_task_internal_record_t      OS_impl_task_table          [OS_MAX_TASKS];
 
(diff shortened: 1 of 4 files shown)


astrogeco:
CCB 2020-06-17: APPROVED

PR #518, opened 2020-06-19
Fix #516, Use absolute timeout for timedlock calls

### LINKED ISSUE #516: Second parameter of pthread_mutex_timedlock should be absolute time not timeout
In src/os/posix/src/os-impl-binsem.c, there are three calls to pthread_mutex_timedlock, each time with &OS_POSIX_BINSEM_MAX_WAIT as the second parameter where OS_POSIX_BINSEM_MAX_WAIT has a value of 2 seconds. because the second parameter of pthread_mutex_timedlock should be an absolute time and is not a timeout, if the mutex is not immediately available, pthread_mutex_timedlock will return with ETIMEDOUT resulting in the OS_BinSemFlush_Impl, OS_GenericBinSemTake_Impl, or OS_BinSemGive_Impl call returning OS_SEM_FAILURE.

Here's a code snippet of the "as is":
    if ( pthread_mutex_timedlock(&sem->id, &OS_POSIX_BINSEM_MAX_WAIT) != 0 )
    {
       return(OS_SEM_FAILURE);
    }

It should look something like:

   struct timespec timeoutTime;
   clock_gettime(CLOCK_REALTIME, &timeoutTime);
   timeoutTime.tv_sec += OS_POSIX_BINSEM_MAX_WAIT.tv_sec;
   timeoutTime.tv_nsec += OS_POSIX_BINSEM_MAX_WAIT.tv_nsec;
   if ( pthread_mutex_timedlock(&sem->id, &timeoutTime) != 0 )
   {
      return(OS_SEM_FAILURE);
   }

Reproducing this issue is tricky as the cFS application has to run long enough to have a mutex not immediately available for the lock. The behavior I'm seeing in my installation is the background task terminates and Ctrl-C can no longer be used to terminate the application. I'm running in a Centos 7 Virtual Box machine.

Jonathan C. Brandenburg
METECS
jonathan.c.brandenburg@nasa.gov

jphickey:
**Describe the contribution**
The pthread API defines the timeout parameter for the pthread_mutex_timedlock call as an absolute value based on CLOCK_REALTIME.

This introduces a wrapper function to calculate the absolute timeout for this.

Fixes #516 

**Testing performed**
Build and execute CFE, sanity check CFE operation, and run all unit tests.
Coverage is not affected because there is no coverage test for POSIX.

**Expected behavior changes**
The timed lock in binary semaphores should now correctly wait for up to 2 seconds for the lock to become available before failing.  The net effect of passing a relative timeout in place of an absolute one meant that the code would not wait, basically turning into a "trylock" instead.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
Unfortunately this wasn't previously noticed during testing because it requires getting two tasks to operate on the same binary sem at exactly the same time, which is difficult to replicate.

### DIFF
--- a/src/os/posix/src/os-impl-binsem.c
+++ b/src/os/posix/src/os-impl-binsem.c
@@ -41,15 +41,35 @@
  * not be relevant in a normally operating system.  This only prevents a
  * deadlock condition in off-nominal circumstances.
  */
-static const struct timespec OS_POSIX_BINSEM_MAX_WAIT =
-{
-        .tv_sec = 2,
-        .tv_nsec = 0
-};
+#define  OS_POSIX_BINSEM_MAX_WAIT_SECONDS       2
 
 
 /* Tables where the OS object information is stored */
 OS_impl_binsem_internal_record_t    OS_impl_bin_sem_table       [OS_MAX_BIN_SEMAPHORES];
+
+/*---------------------------------------------------------------------------------------
+ * Helper function for acquiring the mutex when beginning a binary sem operation
+ * This uses timedlock to avoid waiting forever, and is put into a wrapper function
+ * to avoid pending forever.  The code should never pend on these for a long time.
+ ----------------------------------------------------------------------------------------*/
+int32 OS_Posix_BinSemAcquireMutex(pthread_mutex_t *mut)
+{
+    struct timespec timeout;
+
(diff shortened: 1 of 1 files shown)


jphickey:
@skliper and @astrogeco  - this might be considered for fast track or at least I'd recommend getting this into the branch for the upcoming bundle release in addition to "master" branch.

PR #528, opened 2020-06-29
Fix #525, ensure POSIX stack size meets requirements

### LINKED ISSUE #525: OS_API_Init() failure due to 0 stack size
**Describe the bug**
OS_API_Init() fails on generic-linux due to a stack size of 0 being used for the console task.

**To Reproduce**
1. Build the provided example using the 'generic-linux' BSP.
2. Execute the provided example.

**Expected behavior**
The three test tasks should execute.

**Actual behavior**
OS_API_Init() fails with the following error message (debug messages enabled):

`OS_Posix_InternalTaskCreate_Impl():473:pthread_attr_setstacksize error in OS_TaskCreate: Invalid argument
`

**Code snips**
The error occurs on the following call to pthread stack size in OS_Posix_InternalTaskCreate_Impl():

`return_code = pthread_attr_setstacksize(&custom_attr, stacksz);
`

The reason it fails is because the stacksz is set to zero in OS_ConsoleCreate_Impl():

`return_code = OS_Posix_InternalTaskCreate_Impl(&consoletask, OS_CONSOLE_TASK_PRIORITY, 0,
                    OS_ConsoleTask_Entry, local_arg.opaque_arg);
`

**System observed on:**
 - Hardware: Dell Precision 7540 Laptop
 - OS: WSL2 Ubuntu 18.04.4 LTS
 - Versions: OSAL master 

**Additional context**
This issue is resolved by using a stack size of PTHREAD_STACK_MIN instead of 0:

`return_code = OS_Posix_InternalTaskCreate_Impl(&consoletask, OS_CONSOLE_TASK_PRIORITY, PTHREAD_STACK_MIN,
                    OS_ConsoleTask_Entry, local_arg.opaque_arg);
`

**Reporter Info**
Adam St. Amand

jphickey:
**Describe the contribution**
The `pthread_attr_setstacksize()` function stipulates that it may fail if the user-supplied stack size is not at least
`PTHREAD_STACK_MIN` and also possibly a multiple of page size.

This partially reverts previous PR #508 and adds back rounding up to `PTHREAD_STACK_MIN` and also adds rounding up to a system page size.  However the check for zero stack still remains at the shared level so attempts to create a task with zero
stack will still fail.  This allows internal helper threads to be created with a default minimum stack size, however.

Fixes #525 

**Testing performed**
Build and sanity check CFE executing as both normal user and root user.
Execute all unit tests as both normal user and root user.

**Expected behavior changes**
No failure due to stack size not meeting system requirements.

**System(s) tested on**
Ubuntu 20.04, running as normal user.
CentOS 8.1 in VM running as root user.

**Additional context**
This also moves the `pthread_attr_setstacksize()` call such that it is configured regardless of whether task priorities are enabled nor not (i.e. non root user).  The normal user should still be able to configure the stack size even if they lack permission to set scheduling priority.

Also confirmed that by moving the call to `pthread_attr_setstacksize()` that the original bug report is reproducible in a non-root environment - which is also fixed by this change.

### DIFF
--- a/src/os/posix/src/os-impl-tasks.c
+++ b/src/os/posix/src/os-impl-tasks.c
@@ -38,6 +38,13 @@
 #include "os-shared-task.h"
 #include "os-shared-idmap.h"
 
+/*
+ * Defines
+ */
+#ifndef PTHREAD_STACK_MIN
+#define PTHREAD_STACK_MIN   (8*1024)
+#endif
+
 /* Tables where the OS object information is stored */
 OS_impl_task_internal_record_t      OS_impl_task_table          [OS_MAX_TASKS];
 
@@ -416,6 +423,8 @@ int32 OS_Posix_TaskAPI_Impl_Init(void)
    }
 #endif
 
+   POSIX_GlobalVars.PageSize = sysconf(_SC_PAGESIZE);
+
    return OS_SUCCESS;
 } /* end OS_Posix_TaskAPI_Impl_Init */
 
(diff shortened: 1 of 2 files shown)


src/os/posix/src/os-impl-tasks.c
```diff
@@ -416,6 +424,24 @@ int32 OS_Posix_TaskAPI_Impl_Init(void)
...
+       /*
+        * convert to a mask.  Should already be a power of 2,
+        * but if not, make it so.
+        */
+       --pagesz;
+       pagesz |= (pagesz >> 1);
+       pagesz |= (pagesz >> 2);
+       pagesz |= (pagesz >> 4);
+       pagesz |= (pagesz >> 8);
+       pagesz |= (pagesz >> 16);
```
skliper, line 439: But if it's not a power of two, won't it possibly fail the "may need to be a multiple" condition?  Why not leave as is so it won't possibly fail (and update the math to match)?

mlouielu:
Should the travis CI also need to run with non-root/root user?

src/os/posix/src/os-impl-tasks.c
mlouielu: Also, as the manpage, it said that it will only raise `EINVAL` on some system. I test it on my x86_64 i7 with `16385` (_SC_PAGESIZE = 4096), it didn't raise any error number to me.

jphickey: > But if it's not a power of two, won't it possibly fail the "may need to be a multiple" condition? Why not leave as is so it won't possibly fail (and update the math to match)?

Good point, shouldn't force to power of two.  Commit 43c48b2 changes to just making it a multiple of page size, whatever the number may be.

jphickey: >  as the manpage, it said that it will only raise `EINVAL` on some system

Right, and because we don't know whether or not the host system does or does not need this, we should use whole pages just in case.  On a virtual memory/MMU platform internally the memory allocation is almost certainly done in whole pages anyway.

skliper:
> ~Should the travis CI also need to run with non-root/root user?~ (Ah, no need to do that, this PR do `pthread_attr_setstacksize` regardless of the `EnableTaskPriorities`

Note I did add nasa/cFS#107 for a CI privileged run for better test coverage.

astrogeco:
CCB 202-07-01: APPROVED

PR #533, opened 2020-07-06
Fix #532, Set pthread names to match CFE tasks names

### LINKED ISSUE #532: Set posix thread names using task name when creating CFE tasks
It would be helpful for the pthread name to match the CFE task name in the underlying OS on POSIX platforms.  This would allow for a better user experience when debugging CFE applications using Linux based utilities i.e. strace, htop, etc.

This can be accomplished by using the _pthread_setname_np_ function in the _OS_Posix_InternalTaskCreate_Impl_ function.  Since this is a non-posix function, each BSP can specify the availability of this function through a #define in osconfig.h

None

**Additional context**
None

**Requester Info**
Dan Burns (GSFC - 596.0)
daniel.s.burns@nasa.gov

dsburns:
**Describe the contribution**
This change allows underlying OS tools to view thread names for platforms that support the pthread_setname_np function.

**Testing performed**
1. Set #define OS_HAVE_PTHREAD_SETNAME_NP in osconfig for target
2. Run CFE environment
3. Run htop under Linux and verify thread names appear

**Expected behavior changes**
 - API Change: Adds a task_name parameter to the OS_Posix_InternalTaskCreate_Impl function
 - Behavior Change: None

**System(s) tested on**
 - Hardware: PC
 - OS: Ubuntu 18.04
 - Versions: cFE 6.7.0, OSAL 5.0.0

**Additional context**
None

**Third party code**
None

Daniel Burns
GSFC - Code 596.0
daniel.s.burns@nasa.gov

### DIFF
--- a/src/bsp/pc-linux/config/osconfig.h
+++ b/src/bsp/pc-linux/config/osconfig.h
@@ -196,4 +196,10 @@
  */
 #define OSAL_DEBUG_PERMISSIVE_MODE
 
+/*
+ * If OS_HAVE_PTHREAD_SETNAME_NP is defined, this will allow underlying pthread names
+ * to match CFE task names
+ */
+#define OS_HAVE_PTHREAD_SETNAME_NP
+
 #endif
--- a/src/os/posix/os-posix.h
+++ b/src/os/posix/os-posix.h
@@ -107,7 +107,4 @@ int32 OS_Posix_StreamAPI_Impl_Init(void);
 int32 OS_Posix_DirAPI_Impl_Init(void);
 int32 OS_Posix_FileSysAPI_Impl_Init(void);
 
-int32 OS_Posix_InternalTaskCreate_Impl (pthread_t *thr, uint32 priority, size_t stacksz, PthreadFuncPtr_t Entry, void *entry_arg);
-
-
-
+int32 OS_Posix_InternalTaskCreate_Impl (pthread_t *thr, const char *thread_name, uint32 priority, size_t stacksz, PthreadFuncPtr_t Entry, void *entry_arg);
--- a/src/os/posix/osapi.c
+++ b/src/os/posix/osapi.c
@@ -163,6 +163,10 @@ const OS_ErrorTable_Entry_t OS_IMPL_ERROR_NAME_TABLE[] = { { 0, NULL } };
 static void  OS_CompAbsDelayTime( uint32 milli_second , struct timespec * tm);
 static int   OS_PriorityRemap(uint32 InputPri);
 
+#ifdef OS_HAVE_PTHREAD_SETNAME_NP
+extern int pthread_setname_np(pthread_t thread, const char *name);
+#endif
+
 
 /*----------------------------------------------------------------
  *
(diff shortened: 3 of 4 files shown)


CDKnightNASA:
A lot of other OSAL "objects" (mutex, semaphores, etc. For example, see https://github.com/nasa/osal/blob/master/src/os/shared/src/osapi-mutex.c ). These all follow the same basic pattern, and include names and can map from name to ID and back (at the OSAL layer.) Would it be worthwhile to use/support this same architecture for tasks, and reflect the names down to the OS for those that support naming but otherwise have name stored in OSAL?

dsburns:
The _OS_TaskCreate_ function already takes a _task_name_ parameter.  The parameter is copied to the _task_name_ field of the _OS_task_internal_record_t_ struct for the associated task ID.  It appears this implementation already matches the mutex/sempahore implementation (unless i misinterpreted your comment).  The proposed solution leverages the already existing _task_name_ field for the OSAL task to set the pthread name in OS’s that support this function.

skliper:
What about a module/plug-in?  Not in the core code (or even necessary as part of the framework), but user's who want it could add it in.

jphickey:
> What about a module/plug-in?

I was thinking it could be useful in a generic sense to allow the BSP to register a callback function to implement additional logic on certain ops.  Then the BSP can invoke the non-posix function, which would be fine.

astrogeco:
For more info, I asked Daniel's to submit the PR given that he implemented this for a mission and there is a valid usecase.

jphickey:
I'm not saying its not "nice to have" -- it does seem useful, but it is non-standard/non-posix.  I'm fine with a solution that confines the non-posix aspects to a separate file in the BSP.

FWIW, we might have to do something similar for thread CPU affinity so its worthwhile to come up with a proper way to isolate this stuff not just conditional compile which is hard to maintain.

acudmore:
This is a tough one.. I do see the usefulness of it, but I do agree with Joe, he has worked hard to get non-standard code and ifdefs out of the OSAL implementations.
The smallsat group needs to upgrade to the cFE 6.8 and the corresponding OSAL in the future for the bug fixes and improvements to Linux exception handling, so I would suggest the following:
- keep the patch in the SmallSat repo for now, since it provides value to the projects
- the community can work on a more modular solution that we can also use to implement the CPU affinity code that Joe mentioned
- when the smallsat group upgrades to the next cFE/OSAL hopefully we will have the solution implemented so a patch does not need to be re-applied.

dsburns:
If you guys provide a bit of direction on the callback architecture, I'm willing to put something together in my free time.

jphickey:
FYI - see issue #540 and PR #541 for my suggestions.

I think this general structure would work for several purposes.  In relation to this ticket, it is fairly easy/trivial to call `pthread_setname_np` in the event handler for `OS_EVENT_TASK_STARTUP`.  The handler would reside in the CFE PSP for pc-linux, and OSAL doesn't really have any part of it, other than to invoke the event notification at the right time.

skliper:
@dsburns is the #541 approach acceptable?  If so I'd like to close this issue.

dsburns:
Looks good to me.

astrogeco:
> Looks good to me.

great, I'll close this PR then. Thanks again for getting this started!

PR #539, opened 2020-07-09
Fix #538, correct loop variable for coverage install

### LINKED ISSUE #538: Incorrect install/staging location for coverage test binaries
**Describe the bug**
Typo in cmake file - the loop variable when installing coverage binaries is `TGT` but the DESTINATION refers to `TGTNAME`.

**To Reproduce**
Build with multiple CPUs sharing the same target+platform config.

**Expected behavior**
Binaries should be installed for everything in INSTALL_TARGET_LIST.

**Code snips**
https://github.com/nasa/osal/blob/80fdd67a6c28a962164697ece12570142331ea44/src/unit-test-coverage/CMakeLists.txt#L86-L89

**System observed on:**
Ubuntu 20.04

**Additional context**
The CFE scripts also do set `TGTNAME` in other places so it by chance had a leftover value that happened to be the same thing, and therefore this wouldn't be noticed when building for a single target (the default case).

**Reporter Info**

jphickey:
**Describe the contribution**
Install use the loop variable (TGT) not TGTNAME for DESTINATION.

Fixes #538 

**Testing performed**
Build with ENABLE_UNIT_TESTS=TRUE and a configuration with multiple CPUs sharing a single arch+platform config.

**Expected behavior changes**
Without this fix, the coverage binaries are installed twice to CPU2 but not at all for CPU1.
With this fix, they are correctly installed for CPU1 and CPU2.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
No change to behavior at all - just an install path fix.

### DIFF
--- a/src/unit-test-coverage/CMakeLists.txt
+++ b/src/unit-test-coverage/CMakeLists.txt
@@ -84,7 +84,7 @@ function (add_coverage_testrunner TESTNAME FSW_SRCFILE TESTCASE_SRCFILE)
     add_test(${TESTNAME} ${TESTNAME}-testrunner)
 
     foreach(TGT ${INSTALL_TARGET_LIST})
-        install(TARGETS ${TESTNAME}-testrunner DESTINATION ${TGTNAME}/${UT_INSTALL_SUBDIR})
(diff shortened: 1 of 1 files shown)


astrogeco:
CCB-20200715 APPROVED

PR #541, opened 2020-07-10
Fix #540, add event callback framework

### LINKED ISSUE #540: OSAL event callback framework for platform-specific handling
Many platforms support extra non-standardized features, but OSAL cannot make use of these.  This includes items such as setting the processor affinity in a multi-processor system, or propagating the user-friendly task name at the OS level, as was suggested in #532. 

An event callback framework could help solve this problem, allowing those platform-specific features to be invoked from the PSP/BSP layer while keeping OSAL itself standards-compliant.

Use `#ifdef` conditional compiles for platform-specific features (ugly and not as maintainable).

**Requester Info**

jphickey:
**Describe the contribution**
Adds an event callback mechanism to certain state changes in OSAL.  This allows the CFE PSP to be notified at these points, and therefore it can add platform-specific functionality.  This can, for instance, set the task name as requested in #532 or set the processor affinity in a multi-core setup (TBD).

Fixes #540 

**Testing performed**
Create an event handler in the pc-linux CFE PSP and register it, and print out each event received.  Confirm events are generated as expected, and that it didn't break the FSW in any way.

Note - Not completely tested yet -- this is mainly a proof of concept and pushed for design review at this time.

**Expected behavior changes**
Adds ability to implement custom platform-specific functionality for defined events.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
Previous discussion in #532.

### DIFF
--- a/src/os/inc/osapi-os-core.h
+++ b/src/os/inc/osapi-os-core.h
@@ -177,6 +177,75 @@ typedef enum
 } OS_StreamState_t;
 
 /**
+ * @brief A set of events that can be used with event callback routines
+ */
+typedef enum
+{
+    OS_EVENT_RESERVED = 0,          /**< no-op/reserved event id value */
+
+    /**
+     * resource/id has been newly allocated but not yet created.
+     *
+     * This event is invoked from WITHIN the locked region, in
+     * the context of the task which is allocating the resource.
+     *
+     * If the handler returns non-success, the error will be returned
+     * to the caller and the creation process is aborted.
+     */
+    OS_EVENT_RESOURCE_ALLOCATED,
+
+    /**
+     * resource/id has been fully created/finalized.
+     *
+     * Invoked outside locked region, in the context
+     * of the task which created the resource.
+     *
+     * Data object is not used, passed as NULL.
+     *
+     * Return value is ignored - this is for information purposes only.
+     */
+    OS_EVENT_RESOURCE_CREATED,
+
+    /**
+     * resource/id has been deleted.
+     *
+     * Invoked outside locked region, in the context
+     * of the task which deleted the resource.
+     *
+     * Data object is not used, passed as NULL.
+     *
+     * Return value is ignored - this is for information purposes only.
+     */
+    OS_EVENT_RESOURCE_DELETED,
+
+    /**
+     * New task is starting.
+     *
+     * Invoked outside locked region, in the context
+     * of the task which is currently starting, before
+     * the entry point is called.
+     *
+     * Data object is not used, passed as NULL.
+     *
+     * If the handler returns non-success, task startup is aborted
+     * and the entry point is not called.
+     */
+    OS_EVENT_TASK_STARTUP,
+
+    OS_EVENT_MAX                    /**< placeholder for end of enum, not used */
+} OS_Event_t;
+
+/**
+ * @brief A callback routine for event handling.
+ *
+ * @param[in]    event      The event that occurred
+ * @param[in]    object_id  The associated object_id, or 0 if not associated with an object
+ * @param[inout] data       An abstract data/context object associated with the event, or NULL.
+ * @return status Execution status, see @ref OSReturnCodes.
+ */
+typedef int32 (*OS_EventHandler_t)(OS_Event_t event, osal_id_t object_id, void *data);
+
+/**
  * @brief For the @ref OS_GetErrorName() function, to ensure
  * everyone is making an array of the same length.
  *
@@ -504,6 +573,25 @@ void OS_ForEachObject           (osal_id_t creator_id, OS_ArgCallback_t callback
  * @param[in]  callback_arg Opaque Argument to pass to callback function
   */
 void OS_ForEachObjectOfType     (uint32 objtype, osal_id_t creator_id, OS_ArgCallback_t callback_ptr, void *callback_arg);
+
+/*-------------------------------------------------------------------------------------*/
+/**
+ * @brief Callback routine registration
+ *
+ * This hook enables the application code to perform extra platform-specific
+ * operations on various system events such as resource creation/deletion.
+ *
+ * @note Some events are invoked while the resource is "locked" and therefore
+ * application-defined handlers for these events should not block or attempt
+ * to access other OSAL resources.
+ *
+ * @param[in] handler The application-provided event handler
+ * @return Execution status, see @ref OSReturnCodes.
+ * @retval #OS_SUCCESS @copybrief OS_SUCCESS
+ * @retval #OS_ERROR   @copybrief OS_ERROR
+ */
+int32 OS_RegisterEventHandler (OS_EventHandler_t handler);
+
 /**@}*/
 
 
--- a/src/os/shared/inc/os-shared-common.h
+++ b/src/os/shared/inc/os-shared-common.h
@@ -57,6 +57,12 @@ struct OS_shared_global_vars
    int32             MicroSecPerTick;
    int32             TicksPerSecond;
 
+   /*
+    * The event handler is an application-defined callback
+    * that gets invoked as resources are created/configured/deleted.
+    */
+   OS_EventHandler_t EventHandler;
+
 #ifdef OSAL_CONFIG_DEBUG_PRINTF
    uint8             DebugLevel;
 #endif
@@ -68,6 +74,16 @@ struct OS_shared_global_vars
  * Shared data structure for global values
  */
 extern OS_SharedGlobalVars_t OS_SharedGlobalVars;
+
+/*---------------------------------------------------------------------------------------
+   Name: OS_NotifyEvent
+
+   Purpose: Notify the user application of a change in the state of an OSAL resource
+
+   returns: OS_SUCCESS on success, or relevant error code
+---------------------------------------------------------------------------------------*/
+int32 OS_NotifyEvent(OS_Event_t event, osal_id_t object_id, void *data);
+
 
 /*---------------------------------------------------------------------------------------
    Name: OS_API_Impl_Init
--- a/src/os/shared/src/osapi-common.c
+++ b/src/os/shared/src/osapi-common.c
@@ -65,10 +65,35 @@ OS_SharedGlobalVars_t OS_SharedGlobalVars =
             .ShutdownFlag = 0,
             .MicroSecPerTick = 0, /* invalid, _must_ be set by implementation init */
             .TicksPerSecond = 0,  /* invalid, _must_ be set by implementation init */
+            .EventHandler = NULL,
 #if defined(OSAL_CONFIG_DEBUG_PRINTF)
             .DebugLevel = 1,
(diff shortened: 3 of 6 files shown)


jphickey:
marked as "ccb ready" for concept/architecture discussion (not for merge yet).

src/os/inc/osapi-os-core.h
```diff
@@ -173,6 +173,75 @@ typedef enum
...
+
+/**
+ * @brief A callback routine for event handling.
+ *
+ * @param[in]    event      The event that occurred
+ * @param[in]    object_id  The associated object_id, or 0 if not associated with an object
+ * @param[inout] data       An abstract data/context object associated with the event, or NULL.
+ * @return status Execution status, see @ref OSReturnCodes.
+ */
+typedef int32 (*OS_EventHandler_t)(OS_Event_t event, uint32 object_id, void *data);
```
CDKnightNASA, line 243: confused by the flow of the "data" void *. Usu. with a callback architecture, the opaque data is passed in by the registration and passed through as another "in" to the callback function pointer. But the registration does not define the opaque and this is defined as in/out? (Which implies that the caller would want/need to understand the contents of the data buffer.)

CDKnightNASA: (may be a good opportunity to draw up a sequence diagram or some other way to describe the flow of control and data handoffs.)

CDKnightNASA:
Also would like consideration how this might play with the OSAL "object id" framework. (And currently threads/tasks do not use the object id framework...which it should!) It might make sense that all "objects" would have this capability.

astrogeco:
CCB-2020-07-15: Good architecture, conversation about whether this is the appropriate way to solve the request in https://github.com/nasa/osal/pull/533

astrogeco:
@acudmore and @jwilmot thoughts?

jphickey:
Update - this PR here has some cleanup around the deletion of objects that would be nice to have for other reasons.  IIRC it was discussed/reviewed in CCB a while back and there were no objections.

@astrogeco - Can this be merged?  

If still on the fence about the callback structure, then if ncessary I can split it into a separate PR which consolidates the deletion/cleanup code (which I would like to have merged) and the callback bits.

astrogeco:
Did we want to get a review by either of the architects?

skliper:
Maybe @jwilmot  and @klystron78.  They have a potential to take advantage of this for the SMP work.

astrogeco:
@jphickey let's split this into two PRs so we can accelerate merging the callback stuff.

src/os/inc/osapi-os-core.h
```diff
@@ -377,6 +445,26 @@ void OS_ForEachObject           (uint32 creator_id, OS_ArgCallback_t callback_pt
...
+ * @note Some events are invoked while the resource is "locked" and therefore
+ * application-defined handlers for these events should not block or attempt
+ * to access other OSAL resources.
+ *
+ * @param[in] handler The application-provided event handler
+ * @return Execution status, see @ref OSReturnCodes.
+ * @retval #OS_SUCCESS @copybrief OS_SUCCESS
+ * @retval #OS_ERROR   @copybrief OS_ERROR
+ */
+int32 OS_RegisterEventHandler (OS_EventHandler_t handler);
```
klystron78, line 464: It might make sense to have a bitmask so the caller can specify which events are of interest. This idea also scales forward as more events are added. It could be 32-bit or 64-bit number which gives that many extra bits for future use. Perhaps 0 means "all events."

src/os/shared/inc/os-shared-common.h
```diff
@@ -57,6 +57,12 @@ struct OS_shared_global_vars
    int32             MicroSecPerTick;
    int32             TicksPerSecond;
 
+   /*
+    * The event handler is an application-defined callback
+    * that gets invoked as resources are created/configured/deleted.
+    */
+   OS_EventHandler_t EventHandler;
```
klystron78, line 64: Can there only be one? It might be useful to follow the observer/observable pattern and have a list of observers, so that apps can monitor events as well to get an idea of the state of the system. I am not fully briefed on the use-case for this work, so this thought may be beyond scope.

src/os/shared/src/osapi-binsem.c
```diff
@@ -157,14 +157,8 @@ int32 OS_BinSemDelete (uint32 sem_id)
    {
       return_code = OS_BinSemDelete_Impl(local_id);
 
-      /* Free the entry in the master table now while still locked */
-      if (return_code == OS_SUCCESS)
```
klystron78, line 161: return_code is set on line 158 and not checked with this deletion. It's set again on 161.

src/os/shared/src/osapi-common.c
```diff
@@ -199,6 +224,24 @@ int32 OS_API_Init(void)
...
+ *
+ * Function: OS_RegisterEventHandler
+ *
+ *  Purpose: Implemented per public OSAL API
+ *           See description in API and header file for detail
+ *
+ *-----------------------------------------------------------------*/
+int32 OS_RegisterEventHandler (OS_EventHandler_t handler)
+{
+    if (handler == NULL)
```
klystron78, line 237: Is it an error if null is passed in? The effect would be to say "I am no longer interested in notifications," which seems like would not be an error.

src/os/shared/src/osapi-countsem.c
```diff
@@ -150,14 +150,8 @@ int32 OS_CountSemDelete (uint32 sem_id)
    {
       return_code = OS_CountSemDelete_Impl(local_id);
```
klystron78, line 151: Like before, return_code is set and not checked. It's set again on 154.

src/os/shared/src/osapi-dir.c
```diff
@@ -187,14 +187,8 @@ int32 OS_DirectoryClose(uint32 dir_id)
    {
        return_code = OS_DirClose_Impl(local_id);
```
klystron78, line 188: same as above

src/os/shared/src/osapi-file.c
```diff
@@ -223,14 +223,8 @@ int32 OS_close (uint32 filedes)
     {
         return_code = OS_GenericClose_Impl(local_id);
```
klystron78, line 224: same as above

src/os/inc/osapi-os-core.h
klystron78: Chris has a good point. This should be some context registered with the handler and passed back to it. The concept you're talking about with using it for extra event parameters is going to lead to you having to either force the user to typecast in their callback or create a set of API functions to make the opaque data useful.

src/os/shared/src/osapi-module.c
```diff
@@ -289,14 +289,8 @@ int32 OS_ModuleUnload ( uint32 module_id )
          */
         return_code = OS_ModuleUnload_Impl(local_id);
```
klystron78, line 290: same thing with return_code set and then not checked, and set again below

src/os/shared/src/osapi-mutex.c
```diff
@@ -145,14 +145,8 @@ int32 OS_MutSemDelete (uint32 sem_id)
    {
       return_code = OS_MutSemDelete_Impl(local_id);
```
klystron78, line 146: same thing with return_code set and then not checked, and set again below

src/os/shared/src/osapi-queue.c
```diff
@@ -156,14 +156,8 @@ int32 OS_QueueDelete (uint32 queue_id)
    {
       return_code = OS_QueueDelete_Impl(local_id);
```
klystron78, line 157: same thing with return_code set and then not checked, and set again below

src/os/shared/src/osapi-task.c
```diff
@@ -269,24 +277,14 @@ int32 OS_TaskDelete (uint32 task_id)
...
+      return_code = OS_ObjectIdFinalizeDelete(return_code, record);
    }
 
    /*
    ** Call the thread Delete hook if there is one.
    */
-   if (delete_hook != NULL)
+   if (return_code == OS_SUCCESS && delete_hook != NULL)
    {
       delete_hook();
```
klystron78, line 289: Do we need both delete hook and event for deletion?

src/os/shared/src/osapi-timebase.c
```diff
@@ -260,14 +260,8 @@ int32 OS_TimeBaseDelete(uint32 timer_id)
     {
         return_code = OS_TimeBaseDelete_Impl(local_id);
```
klystron78, line 261: same thing with return_code set and then not checked, and set again below

src/os/shared/src/osapi-binsem.c
jphickey: The return code is checked inside the `OS_ObjectIdFinalizeDelete` function.  There is cleanup to do for both success and failure cases.  (This is a mirror image of the way the allocation/creation works too).

jphickey:
See split in PRs #566, #567 

Review of this should focus on commit c0bb696.  Converted back to draft while discussing the event callback API and use case.

jphickey:
Note to reviewers - the initial use case for this callback is in #532, which requires invoking a Linux-specific API to set the name of the task.

It is _not_ intended for dynamic application level use - it is intended more as an extension mechanism to replace what would otherwise be platform-specific `#ifdef` routines.  As such, it is a simpler problem - there should only ever be one callback function registered, which should be done immediately after `OS_API_Init()`, and it should never be subsequently un-registered.

If we wanted to provide a more generic application-level callback - then yes I agree with the comments regarding providing a method to subscribe to only certain events, and providing another application-provided object to pair with the callback, and so forth.  But that wasn't really the objective here.  We can certainly do that if folks think its useful, but I don't really see the use-case for it right now.  The singlular BSP/PSP-provided callback proposed here is much simpler to implement, as any dynamic registration/unregistration would be much more involved i.e. it would have to maintain a list, proper locking to support being invoked by any thread, etc.

skliper:
Need a path to closure, marking for review (unless we can reach consensus before CCB).

astrogeco:
[CCB 2020-09-23](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-23-2020) rebase on main and revisit on 2020-09-30. @acudmore and @CDKnightNASA will examine

jphickey:
Rebased to latest main branch.  Ready for (re-)review.

Also see the use-case example at https://github.com/nasa/PSP/compare/main...jphickey:jph-wip-linux-event-handler

I have confirmed that when running this on my linux box with the change above applied, then the `/proc` filesystem reads accordingly, i.e. the `/proc/<pid>/task/<threadid>/comm` special files reads as `CFE_EVS`, `CFE_SB`, etc.

astrogeco:
[CCB 2020-09-30](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-30-2020): APPROVED

PR #566, opened 2020-08-18
Fix #565, propagate return code from OS_TaskRegister_Impl()

### LINKED ISSUE #565: OS_TaskPrepare should propagate return code.
**Describe the bug**
In the internal `OS_TaskPrepare()` routine, it calls a low-level implementation routine but does not propagate the return code.

**To Reproduce**
More theoretical than real.  In POSIX this depends on the underlying `pthread_setspecific()` routine returning an error, of which the only documented error (at least per POSIX) is `ENOMEM`.  However that doesn't mean that there couldn't be other implementation-specific errors on other systems.

**Expected behavior**
If the `OS_TaskRegister_Impl()` fails, this result should be propagated up the stack and prevent the task from starting.

**Code snips**
See here:
https://github.com/nasa/osal/blob/8cfd6fe71a5506be8e463f26d92441785fd3e242/src/os/shared/src/osapi-task.c#L117-L125

**System observed on:**

**Additional context**
If this fails then it means dependent routines like `OS_TaskGetId()` will also be broken, so it really should block startup of the task and return the error to the caller.  Catching this type of error early, when possible, is usually better than getting an obfuscated error later.

**Reporter Info**

jphickey:
**Describe the contribution**
If this routine fails then return the error to the caller, which will also prevent the task from starting.

Fixes #565 

**Testing performed**
Build and sanity check CFE
Build and run all unit tests incl. coverage.

**Expected behavior changes**
No impact to behavior, as no known way to force this error, but will correctly propagate an error if one were to occur.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/src/os/shared/src/osapi-task.c
+++ b/src/os/shared/src/osapi-task.c
@@ -116,9 +116,10 @@ static int32 OS_TaskPrepare(uint32 task_id, osal_task_entry *entrypt)
 
    if (return_code == OS_SUCCESS)
    {
-      OS_TaskRegister_Impl(task_id);
+       return_code = OS_TaskRegister_Impl(task_id);
    }
-   else
+
+   if (return_code != OS_SUCCESS)
    {
       *entrypt = NULL;
    }


astrogeco:
[CCB 2020-08-19](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-August-19-2020) APPROVED

PR #567, opened 2020-08-18
Fix #564, consolidate delete finalization code

### LINKED ISSUE #564: Consolidate cleanup code into a separate "FinalizeDelete" routine
All OSAL objects share a basically common pattern for finalization after creation and deletion.  For creation the finalization was consolidated into a `OS_ObjectIdFinalizeNew` routine, but the deletion finalization is basically repeated as it only involved clearing a single global.

Even though deletion is currently trivial, it is beneficial to consolidate this code as it could become less trivial, as in issue #540 

**Additional context**
This was split from #540

This is a necessary prerequisite in order to implement any sort of common/generic handling without repeating the code everywhere.  It also makes deletion more of a mirror-image of creation, which is really the way it should be.

**Requester Info**

jphickey:
**Describe the contribution**

Remove repetitive clearing of the global ID and unlocking global table and replace with common implementation in the
idmap source file.  In addition to removing repeated logic, this makes deletion more like an inverse of creation, and provides
a common/single point where additional deletion-related logic can be added.

Fixes #564 

**Testing performed**
Build and sanity test CFE
Confirm all unit tests

**Expected behavior changes**
No impact to external API or behavior.  Internal code consolidation only.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
See also #540 and #532 

### DIFF
--- a/src/os/shared/inc/os-shared-idmap.h
+++ b/src/os/shared/inc/os-shared-idmap.h
@@ -297,6 +297,19 @@ int32 OS_ObjectIdAllocateNew(uint32 idtype, const char *name, uint32 *array_inde
 int32 OS_ObjectIdFinalizeNew(int32 operation_status, OS_common_record_t *record, uint32 *outid);
 
 /*----------------------------------------------------------------
+   Function: OS_ObjectIdFinalizeDelete
+
+    Purpose: Completes a delete operation
+             If the operation was successful, the OSAL ID is deleted and returned to the pool
+             If the operation was unsuccessful, no operation is performed.
+             The global table is unlocked for future operations
+
+    Returns: OS_SUCCESS on success, or relevant error code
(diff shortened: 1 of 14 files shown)


jphickey:
Note- commit just 2cc7dc5 corrects an unused variable issue reported by cppcheck.

jphickey:
@klystron78 - FYI this is largely the same subject as your previous review of #541.  However note that the return_code is indeed still checked, the objective here is to put the cleanup code inside a common function (in this case `OS_ObjectIdFinalizeDelete`) rather than copying the same check everywhere.

astrogeco:
[CCB 2020-08-19](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-August-19-2020) APPROVED

PR #595, opened 2020-09-09
Fix #589, promote OS_ObjectIdToArrayIndex to public API

### LINKED ISSUE #589: Promote the internal `OS_ObjectIdToArrayIndex` function to the public API
Currently OSAL only provides a public API for `OS_ConvertToArrayIndex` which operates indiscriminately on any object type.  For instance, if passed a task ID, it returns a value in the range of `[0..OS_MAX_TASKS)` and if passed a queue ID it will return a value in the range of `[0..OS_MAX_QUEUES)` and so forth.

But code using this function is using it to index an array of a specific object type (e.g. tasks).  So this creates an opportunity for error if passed the wrong object ID - the caller may get a successful conversion but still get an index that is beyond the range of their local table.

The `OS_ObjectIdToArrayIndex` function is better because it first enforces that the passed-in ID actually refers to the intended object type.  The public API is actually just a wrapper around this that _defeats_ the type check.  For some reason this got relegated to an internal API.

Keep as is, accept risk of getting an index outside the expected range if IDs get mixed.

**Additional context**
As the function already exists it should be just a matter of putting its prototype into the public API, and providing a UT stub for it (easy).

**Requester Info**

jphickey:
**Describe the contribution**

Move this existing function into the public API, as it is performs more verification than the `OS_ConvertToArrayIndex`
function.

Fixes #589 

**Testing performed**
Build and sanity test CFE
Run all tests

**Expected behavior changes**
None. 

**System(s) tested on**
Ubuntu 20.04

**Additional context**
 No change to function, just moved to public API.

### DIFF
--- a/src/os/inc/osapi-os-core.h
+++ b/src/os/inc/osapi-os-core.h
@@ -428,6 +428,12 @@ uint32 OS_IdentifyObject       (osal_id_t object_id);
  * @note This does NOT verify the validity of the ID, that is left to the caller.
  * This is only the conversion logic.
  *
+ * This routine accepts any object type, and returns a value based on the
+ * maximum number of objects for that type.  This is equivalent to invoking
+ * OS_ObjectIdToArrayIndex() with the idtype set to OS_OBJECT_TYPE_UNDEFINED.
+ *
+ * @sa OS_ObjectIdToArrayIndex
+ *
  * @param[in]  object_id    The object ID to operate on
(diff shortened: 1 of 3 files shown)


astrogeco:
[CCB 2020-09-09](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-09-2020) APPROVED, Added Public API

PR #599, opened 2020-09-16
Fix #610, Add vxworks7 support

### LINKED ISSUE #610: Add support for vxWorks 7
cFS traditionally has supported vxWorks 6. There are projects interested in using vxWorks 7 with cFS, so support must be added. This requires work in PSP and OSAL.

An ideal solution is to support both vxWorks 6 and 7 with as little cloning of old vxWorks 6 as possible. On PSP, the current work is to copy mcp750-vxworks for the working platform. (5020) OSAL's goal is to have the vxworks osal support both 6 and 7.

**Additional context**

**Requester Info**
Steven Seeger / Embedded Flight Systmes, Inc. / GSFC Code 582

klystron78:
**Describe the contribution**
This is a more Joe-approved vxworks7 addition without the copy of the code.

**Testing performed**
None yet. I'm going to bed. But I wanted to look at it with the CCB. I will try and do testing before the CCB.

Steven Seeger, Embedded Flight Systems, Inc.

### DIFF
--- a/src/os/vxworks/CMakeLists.txt
+++ b/src/os/vxworks/CMakeLists.txt
@@ -14,7 +14,7 @@ set(VXWORKS_BASE_SRCLIST
     src/os-impl-console.c
     src/os-impl-countsem.c
     src/os-impl-errors.c
-    src/os-impl-dirs.c
+    src/os-impl-dirs-globals.c
     src/os-impl-files.c
     src/os-impl-filesys.c
     src/os-impl-heap.c
@@ -45,6 +45,15 @@ else ()
     )
 endif ()
 
+if (CMAKE_SYSTEM_VERSION VERSION_GREATER_EQUAL 7.0)
+    list(APPEND VXWORKS_IMPL_SRCLIST
+        ../portable/os-impl-posix-dirs.c
+    )
+else ()
+    list(APPEND VXWORKS_IMPL_SRCLIST
+        src/os-impl-dirs.c
+    )
+endif ()
 # If some form of module loading is configured,
 # then build the module loader
 if (OSAL_CONFIG_INCLUDE_DYNAMIC_LOADER)
--- a/src/os/vxworks/src/os-impl-common.c
+++ b/src/os/vxworks/src/os-impl-common.c
@@ -184,8 +184,11 @@ int32 OS_VxWorks_GenericSemTake(SEM_ID vxid, int sys_ticks)
          * check for the timeout condition,
          * which has a different return code and
          * not necessarily an error of concern.
+         *
+         * vxworks7: if sys_ticks == 0, then if the semaphore can
+         * not be taken S_objLib_OBJ_UNAVAILABLE will be returned
          */
-        if (errno == S_objLib_OBJ_TIMEOUT)
+        if ((errno == S_objLib_OBJ_TIMEOUT) || (!sys_ticks && (errno == S_objLib_OBJ_UNAVAILABLE)))
         {
             return OS_SEM_TIMEOUT;
         }
--- a/src/os/vxworks/src/os-impl-dirs-globals.c
+++ b/src/os/vxworks/src/os-impl-dirs-globals.c
@@ -19,36 +19,37 @@
  */
 
 /**
- * \file     os-impl-tasks.h
+ * \file     os-impl-dirs.c
  * \ingroup  vxworks
  * \author   joseph.p.hickey@nasa.gov
  *
  */
 
-#ifndef INCLUDE_OS_IMPL_TASKS_H_
-#define INCLUDE_OS_IMPL_TASKS_H_
+/****************************************************************************************
+                                    INCLUDE FILES
+****************************************************************************************/
 
-#include <osconfig.h>
-#include <taskLib.h>
+#include "os-vxworks.h"
+#include "os-impl-dirs.h"
+#include "os-shared-dir.h"
 
-#if defined(VX_WIND_TCB_SIZE)
-/* vxworks >= 7.0 should provide this symbol via taskLib.h. WIND_TCB is an opaque type */
-typedef char OS_VxWorks_TCB_t[VX_WIND_TCB_SIZE];
-#else
-/* older vxworks expose the definition of VX_WIND_TCB_SIZE */
-typedef WIND_TCB OS_VxWorks_TCB_t;
-#endif /* !defined(VX_WIND_TCB_SIZE) */
 
-/*tasks */
-typedef struct
+/*
+ * The directory handle table.
+ */
+OS_impl_dir_internal_record_t OS_impl_dir_table[OS_MAX_NUM_OPEN_DIRS];
+
+
+/*----------------------------------------------------------------
+ *
+ * Function: OS_VxWorks_DirAPI_Impl_Init
+ *
+ *  Purpose: Local helper routine, not part of OSAL API.
+ *
+ *-----------------------------------------------------------------*/
+int32 OS_VxWorks_DirAPI_Impl_Init(void)
 {
-    OS_VxWorks_TCB_t tcb; /* Must be first */
-    TASK_ID vxid;
-    void    *heap_block;    /* set non-null if the stack was obtained with malloc() */
-    size_t  heap_block_size;
-} OS_impl_task_internal_record_t;
+   memset(OS_impl_dir_table, 0, sizeof(OS_impl_dir_table));
+   return OS_SUCCESS;
+} /* end OS_VxWorks_DirAPI_Impl_Init */
 
-/* Tables where the OS object information is stored */
-extern OS_impl_task_internal_record_t OS_impl_task_table[OS_MAX_TASKS];
-
-#endif /* INCLUDE_OS_IMPL_TASKS_H_ */
--- a/src/os/vxworks/src/os-impl-dirs.c
+++ b/src/os/vxworks/src/os-impl-dirs.c
@@ -33,24 +33,6 @@
 #include "os-impl-dirs.h"
 #include "os-shared-dir.h"
 
-/*
- * The directory handle table.
- */
-OS_impl_dir_internal_record_t OS_impl_dir_table[OS_MAX_NUM_OPEN_DIRS];
-
-/*----------------------------------------------------------------
- *
- * Function: OS_VxWorks_DirAPI_Impl_Init
- *
- *  Purpose: Local helper routine, not part of OSAL API.
- *
- *-----------------------------------------------------------------*/
-int32 OS_VxWorks_DirAPI_Impl_Init(void)
-{
-    memset(OS_impl_dir_table, 0, sizeof(OS_impl_dir_table));
-    return OS_SUCCESS;
-} /* end OS_VxWorks_DirAPI_Impl_Init */
-
 /*----------------------------------------------------------------
  *
  * Function: OS_DirCreate_Impl
--- a/src/os/vxworks/src/os-impl-tasks.c
+++ b/src/os/vxworks/src/os-impl-tasks.c
@@ -238,7 +238,7 @@ int32 OS_TaskCreate_Impl(osal_index_t task_id, uint32 flags)
 #endif
 
     id.id  = OS_global_task_table[task_id].active_id;
-    status = taskInit(&lrec->tcb,                                              /* address of new task's TCB */
+    status = taskInit((WIND_TCB*)&lrec->tcb,                                   /* address of new task's TCB */
                       (char *)OS_global_task_table[task_id].name_entry, vxpri, /* priority of new task */
                       vxflags,                                                 /* task option word */
                       (char *)actualstackbase,                                 /* base of new task's stack */
--- a/src/unit-test-coverage/vxworks/CMakeLists.txt
+++ b/src/unit-test-coverage/vxworks/CMakeLists.txt
@@ -6,6 +6,7 @@ set(VXWORKS_MODULE_LIST
     console
     countsem
     dirs
+    dirs-globals
     files
     filesys
     idmap
--- a/src/unit-test-coverage/vxworks/adaptors/CMakeLists.txt
+++ b/src/unit-test-coverage/vxworks/adaptors/CMakeLists.txt
@@ -13,6 +13,7 @@ add_library(ut-adaptor-${SETNAME} STATIC
     src/ut-adaptor-console.c
     src/ut-adaptor-countsem.c
     src/ut-adaptor-dirs.c
+    src/ut-adaptor-dirtable-stub.c
     src/ut-adaptor-files.c
     src/ut-adaptor-filesys.c
     src/ut-adaptor-idmap.c
--- a/src/unit-test-coverage/vxworks/adaptors/src/ut-adaptor-dirtable-stub.c
+++ b/src/unit-test-coverage/vxworks/adaptors/src/ut-adaptor-dirtable-stub.c
@@ -19,36 +19,18 @@
  */
 
 /**
- * \file     os-impl-tasks.h
- * \ingroup  vxworks
+ * \file     ut-adaptor-filetable-stub.c
+ * \ingroup  adaptors
  * \author   joseph.p.hickey@nasa.gov
  *
  */
+/* pull in the OSAL configuration */
+#include "osconfig.h"
+#include "ut-adaptor-filetable-stub.h"
 
-#ifndef INCLUDE_OS_IMPL_TASKS_H_
-#define INCLUDE_OS_IMPL_TASKS_H_
+#include <os-vxworks.h>
+#include <os-impl-dirs.h>
+#include <os-shared-dir.h>
 
-#include <osconfig.h>
-#include <taskLib.h>
+OS_impl_dir_internal_record_t OS_impl_dir_table[OS_MAX_NUM_OPEN_DIRS];
 
-#if defined(VX_WIND_TCB_SIZE)
-/* vxworks >= 7.0 should provide this symbol via taskLib.h. WIND_TCB is an opaque type */
-typedef char OS_VxWorks_TCB_t[VX_WIND_TCB_SIZE];
-#else
-/* older vxworks expose the definition of VX_WIND_TCB_SIZE */
-typedef WIND_TCB OS_VxWorks_TCB_t;
-#endif /* !defined(VX_WIND_TCB_SIZE) */
-
-/*tasks */
-typedef struct
-{
-    OS_VxWorks_TCB_t tcb; /* Must be first */
-    TASK_ID vxid;
-    void    *heap_block;    /* set non-null if the stack was obtained with malloc() */
-    size_t  heap_block_size;
-} OS_impl_task_internal_record_t;
-
-/* Tables where the OS object information is stored */
-extern OS_impl_task_internal_record_t OS_impl_task_table[OS_MAX_TASKS];
-
-#endif /* INCLUDE_OS_IMPL_TASKS_H_ */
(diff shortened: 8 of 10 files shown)


src/os/vxworks/src/vxworks7/os-impl-common.c
```diff
@@ -0,0 +1,79 @@
...
+    {
+        /*
+         * check for the timeout condition,
+         * which has a different return code and
+         * not necessarily an error of concern.
+         *
+         * vxworks7: if sys_ticks == 0, then if the semaphore can
+         * not be taken S_objLib_OBJ_UNAVAILABLE will be returned
+         */
+        if ((errno == S_objLib_OBJ_TIMEOUT) || (!sys_ticks && (errno == S_objLib_OBJ_UNAVAILABLE)))
```
jphickey, line 68: I'm confused about why this file had to be forked ... doesn't vxworks6 have a `S_objLib_OBJ_UNAVAILABLE` symbol as well?  Wouldn't this change apply to both versions?  I would think this is actually a bug in the 6.9 OSAL.

src/os/vxworks/inc/vxworks7/os-impl-tasks.h
```diff
@@ -0,0 +1,51 @@
...
+#include <osconfig.h>
+#include <taskLib.h>
+
+
+/*tasks */
+typedef struct
+{
+    /* SDS: WIND_TCB was moved to a private taskLib file. There is an API to access fields within. In this case,
+     * OSAL wants to store other data with WIND_TCB in a single allocaiton of memory, so we only need the size. */
+    char tcb[VX_WIND_TCB_SIZE]; /* Must be first */
```
jphickey, line 40: Rather than forking this file - does vxworks 6.9 also have this `VX_WIND_TCB_SIZE` define?  Also do they give any alignment requirements for it?  My only concern with changing a real type to a generic `char[]` type is that you lose alignment.  As the parent struct contains `void*` and `long` it is probably aligned but worth noting.

jphickey (changes requested):
Getting there ... we have to fix the CI build failure which is probably caused by forking files.

I still think this can be done without forking anything except the dir stuff.

jphickey:
BTW - to comment on the process a bit - we first start with an issue before submitting a PR.  Recommend to basically take the source code comment here:

https://github.com/klystron78/osal/blob/378578829317e3c23e518529b55254918357669e/src/os/vxworks/src/vxworks7/os-impl-common.c#L61-L66

And submitting an issue about this first (i.e. "on vxworks7 such and such does not work") then submitting the PR that addresses it.

jphickey:
FYI - The allocator macro I was referring to in CCB can be seen here for counting sems:

https://github.com/nasa/osal/blob/4f7e201f87839e2c2f3793cce6184c52efd6d14b/src/os/vxworks/inc/os-impl-countsem.h#L37

Similar macro exists for binary sem, mutex, etc.  It would seem possible (maybe even probable?) that wind river would offer a similar macro to statically allocate a task TCB?

klystron78:
> FYI - The allocator macro I was referring to in CCB can be seen here for counting sems:
> 
> https://github.com/nasa/osal/blob/4f7e201f87839e2c2f3793cce6184c52efd6d14b/src/os/vxworks/inc/os-impl-countsem.h#L37
> 
> Similar macro exists for binary sem, mutex, etc. It would seem possible (maybe even probable?) that wind river would offer a similar macro to statically allocate a task TCB?

I don't see one. From taskLib.h:

17 29nov18,gws  made windTcb definition private, added windTcb opaque definition,
 18              made TCB inline functions private,
 19              added VX_WIND_TCB_SIZE for VX_TASK, VX_TASK_INITIALIZE (F11200)

VX_TASK declares space for a task and stack, and it uses VX_WIND_TCB_SIZE to size a character array for the tcb.

src/os/vxworks/src/vxworks7/os-impl-common.c
klystron78: i went ahead and  used an #if

jphickey:
Can you check if vxworks 6.9 provides a `VX_WIND_TCB_SIZE` macro?  It would seem it does not, from the sounds of it.

Which is actually (possibly) a good thing -- because you can probably make a local macro:

```
#ifdef VX_WIND_TCB_SIZE /* vxworks7+ provides this abstract size */
typedef char[VX_WIND_TCB_SIZE]  OS_VxWorks_TCB_t;
#else  /* no abstract size, use WIND_TCB directly */
typedef WIND_TCB  OS_VxWorks_TCB_t;
#endif
```

Then just declare the struct based off `OS_VxWorks_TCB_t` ....

src/os/vxworks/src/vxworks7/os-impl-common.c
jphickey: While I like that we now have it common, the more I'm looking at it, the more I'm failing to see why we need to test for `S_objLib_OBJ_UNAVAILABLE` ...

If called with `sys_ticks` as 0, if the semaphore failed to take for any reason, the proper response up to the app should be `OS_SEM_TIMEOUT`... right?  

Why wouldn't this work:
```
        if (sys_ticks == 0 || errno == S_objLib_OBJ_TIMEOUT)
        {
           return OS_SEM_TIMEOUT;
        }
```

skliper:
Any update on this?  We need to wrap up so we can progress with certification activities.

src/os/vxworks/CMakeLists.txt
```diff
@@ -45,6 +45,19 @@ else ()
...
+    )
+    include_directories(${CMAKE_CURRENT_SOURCE_DIR}/inc/vxworks7)
+    add_definitions(-DOSAL_VXWORKS7)
+else ()
+    list(APPEND VXWORKS_BASE_SRCLIST
+        src/os-impl-dirs.c
+    )
+    include_directories(${CMAKE_CURRENT_SOURCE_DIR}/inc/vxworks6)
+    add_definitions(-DOSAL_VXWORKS6)
+endif ()
```
jphickey, line 60: For consistency I recommend appending to the `VXWORKS_IMPL_SRCLIST` not `VXWORKS_BASE_SRCLIST`

Also those "include_directories" don't seem to exist, and the OSAL_VXWORKS6/OSAL_VXWORKS7 define shouldn't be needed (see my other comment).

src/os/vxworks/inc/os-impl-tasks.h
```diff
@@ -35,7 +35,15 @@
...
 {
+#if defined(OSAL_VXWORKS7)
+    /* SDS: WIND_TCB was moved to a private taskLib file. There is an API to access fields within. In this case,
+     * OSAL wants to store other data with WIND_TCB in a single allocaiton of memory, so we only need the size. */
+    char tcb[VX_WIND_TCB_SIZE]; /* Must be first */
+#elif defined(OSAL_VXWORKS6)
     WIND_TCB tcb;           /* Must be first */
+#else
+#error build system must define OSAL_VXWORKS6 or OSAL_VXWORKS7
+#endif
```
jphickey, line 46: Not sure why this was done with a separate macro (OSAL_VXWORKS6/7) vs. checking the macro that VxWorks 7 directly provides?  I'm also not fond of `#ifdef` directly inside struct declarations.

Since going forward future versions of VxWorks are likely to only provide VX_WIND_TCB_SIZE, then what I would prefer is something like:

```
#ifdef VX_WIND_TCB_SIZE
/* VxWorks >= 7.0 should provide this symbol via taskLib.h, WIND_TCB is an abstract type */
typedef char OS_VxWorks_TCB_t[VX_WIND_TCB_SIZE];
#else
/* Older VxWorks provide the definition of WIND_TCB directly */
typedef WIND_TCB OS_VxWorks_TCB_t;
#endif
```

Then the struct itself is simple - use `OS_VxWorks_TCB_t` as the type for the `tcb` member.

NOTE - when building coverage test the CMakeLists.txt inside vxworks dir isn't used - hence why this failed travis CI (neither is defined).  With the approach above it should work for both FSW and UT.

src/os/vxworks/src/os-impl-common.c
```diff
@@ -192,8 +191,11 @@ int32 OS_VxWorks_GenericSemTake(SEM_ID vxid, int sys_ticks)
          * check for the timeout condition,
          * which has a different return code and
          * not necessarily an error of concern.
+         *
+         * vxworks7: if sys_ticks == 0, then if the semaphore can
+         * not be taken S_objLib_OBJ_UNAVAILABLE will be returned
          */
-        if (errno == S_objLib_OBJ_TIMEOUT)
+        if ((errno == S_objLib_OBJ_TIMEOUT) || (!sys_ticks && (errno == S_objLib_OBJ_UNAVAILABLE)))
```
jphickey, line 198: See my previous comment, can this be simplified to just:

`if (sys_ticks == 0 || errno == S_objLib_OBJ_TIMEOUT)`

jphickey (changes requested):
Still got some stuff to resolve here.  Unit test is broken (see Travis-CI) and I think we can further simplify.

src/os/vxworks/src/os-impl-common.c
klystron78: This doesn't work. errno can also be S_objLib_OBJ_ID_ERROR or S_intLib_NOT_ISR_CALLABLE.

src/os/vxworks/inc/os-impl-tasks.h
klystron78: Did you really just say you aren't fond of #ifdef directly inside struct declarations, and then suggest an #ifdef in a struct declaration? :)

src/os/vxworks/CMakeLists.txt
klystron78: Yeah that was a mistake. I will change it.

klystron78: the include_directories were leftover from the first attempt. I have removed them.

src/os/vxworks/inc/os-impl-tasks.h
klystron78: I like your idea better so I will remove the macros. If vxworks8 does it differently we will have to re-visit. :)

klystron78: Is "abstract type" the correct term to use in your comment? I think it should be "opaque type"

klystron78:
I've done a rebase and pushed this for consideration. I've updated with Joe's latest comments.

I don't know anything about travis-ci or what the problem is. This would be a good time to learn. Since I did a rebase and force push it might make sense to close this pull request and open a new one?

src/os/vxworks/inc/os-impl-tasks.h
jphickey: No, the `#ifdef` is not _inside_ the struct declaration in my suggestion - it controls a typedef.  The typedef is then used in the struct - this is OK.  

This pattern means that the member always must exist and always have the same name - it might differ only in its type.  Putting `#ifdef` directly inside the struct means that names and even the existence of the member is uncertain - this is what I don't like about it. 

And yes, the term "opaque" is probably better here ... I do tend to intermix these terms.  C doesn't really do abstract types like C++ does.

src/os/vxworks/src/os-impl-common.c
jphickey: > This doesn't work. errno can also be S_objLib_OBJ_ID_ERROR or S_intLib_NOT_ISR_CALLABLE

But my point is - if sys_ticks is 0, does that detail matter?

Also - related question - Can `S_objLib_OBJ_UNAVAILABLE` get generated for other reasons too, when sys_ticks is nonzero?

jphickey:
> I've done a rebase and pushed this for consideration. I've updated with Joe's latest comments
 
This looks like a straight rebase to me -- I don't see any differences in the way the struct is defined from this commit to the previous one.  Did you push the whole set of updates you've made?  May want to double check.

> I don't know anything about travis-ci or what the problem is. 

For the travis error click the "details" link then click the build that failed - in this case all 4 have failed so just pick one.  Then scroll down the logs to see this build issue:
```
In file included from /home/travis/build/nasa/osal/src/unit-test-coverage/vxworks/adaptors/src/ut-adaptor-tasks.c:33:0:

/home/travis/build/nasa/osal/src/os/vxworks/inc/os-impl-tasks.h:45:2: error: #error build system must define OSAL_VXWORKS6 or OSAL_VXWORKS7

 #error build system must define OSAL_VXWORKS6 or OSAL_VXWORKS7

  ^~~~~
```

If you implement my previous comment it should fix this issue.... so make sure you've pushed everything.

> Since I did a rebase and force push it might make sense to close this pull request and open a new one?

No need - we update branches all the time prior to merge.  Only need a new PR if it was reverted/rejected post merge for some reason.

jphickey:
@klystron78  - Correction - actually the commits in this PR are still showing to be from Sep 20, which may explain why they look the same to me.  Are you sure you pushed?

klystron78:
> @klystron78 - Correction - actually the commits in this PR are still showing to be from Sep 20, which may explain why they look the same to me. Are you sure you pushed?

Yes I pushed. The error in your previous comment was probably due to the fact that I first had a bad push due to mixing up commit order in an interactive rebase locally. I backed that out and pushed.

I am about to make sure all this runs on the target hardware.

src/os/vxworks/src/os-impl-common.c
klystron78: It can still be called with a bad object id or from an ISR context if sys_ticks==0. Should that ever happen? No. But we wouldn't want to return "everything is ok" if it does.

src/os/vxworks/inc/os-impl-tasks.h
klystron78: Sorry I missed the "typedef" I will change it now and push again.

jphickey:
> Yes I pushed.

Github disagrees --- I'm still seeing the original commits and still on the old baseline from Sept.  I even re-fetched this PR to my local repo and see the same.  I also don't see the normal log messages that appear in the PR when you force push that says you updated it.

klystron78:
> > Yes I pushed.
> 
> Github disagrees --- I'm still seeing the original commits and still on the old baseline from Sept. I even re-fetched this PR to my local repo and see the same. I also don't see the normal log messages that appear in the PR when you force push that says you updated it.

Sorry just noticed that the branch for the PR is vxworks7-2. Not rebase/vxworks7. I will update to that branch.

jphickey (changes requested):
Somehow it looks like typecasts got into this one whereas they were not in there before.  Note that OSAL PR #654 (currently in IC) updates a lot of these so it better if we do NOT include them here or else we'll get merge conflicts.

Requesting that the typecast updates to fix warnings at least be a separate PR.... otherwise looks better now.  I will also pull this to my local machine and test it (haven't done yet).

Have you confirmed coverage testing and all on this?

klystron78:
> Somehow it looks like typecasts got into this one whereas they were not in there before. Note that OSAL PR #654 (currently in IC) updates a lot of these so it better if we do NOT include them here or else we'll get merge conflicts.
> 
> Requesting that the typecast updates to fix warnings at least be a separate PR.... otherwise looks better now. I will also pull this to my local machine and test it (haven't done yet).

So it's required that I have the typecasts and changes in here to avoid compiler warnings (which are errors due to Werror=conversion, on by default in vxworks7) so it seemed to me that this was acceptable to have vxworks7 support. I can take all that out (I made a separate conversion branch the vxworks7 work is on top of) but then I'd have to update my toolchain or change the osal's CMakeLists to avoid conversion errors. What do you prefer?

> 
> Have you confirmed coverage testing and all on this?

I have only run this on vxworks7. I don't have access to run on 6.

klystron78:
Ok this still needs some work. I forgot I had Wno-conversion in my toolchain file, so I've missed some typecasts in the conversion branch. Should I just remove that branch from this entirely and not have any typecasts at all? I can leave Wno-conversion in my toolchain file.

jphickey:
> So it's required that I have the typecasts and changes in here to avoid compiler warnings (which are errors due to Werror=conversion, on by default in vxworks7) so it seemed to me that this was acceptable to have vxworks7 support. 

I don't consider a warning option to be in the critical path of getting VxWorks 7 to work like the other stuff is.  Preference is for focused PRs that can be individually reviewed and tested.  

I'm mainly just requesting that this warning resolution stuff be split into a separate PR.  It kinda requires a different review and testing and one can just as easily add `-Wno-conversion` for the time being which is really the way gcc defaults to in the current builds.

CCB can then decide if we want to build with `-Wconversion` and apply it across the board to all supported plaforms if we do.  

Main thing is to be consistent - we should build all platforms with `-Wconversion` or without it and make it explicit - not be dependent on compiler defaults.   It's really just luck/chance that all cross compilers thus far have been gcc based and they do not include `-Wconversion` with `-Wall` like clang does.

Nothing in that should hold up the rest of the VxWorks 7 stuff though.  So hence a separate PR is what I recommend.

> 
> > Have you confirmed coverage testing and all on this?
> 
> I have only run this on vxworks7. I don't have access to run on 6.

Coverage testing runs on the native host.  Build with `SIMULATION=native ENABLE_UNIT_TESTS=true` and then run `make test` and `make lcov`.

jphickey:
> So it's required that I have the typecasts and changes in here to avoid compiler warnings (which are errors due to Werror=conversion, on by default in vxworks7) so it seemed to me that this was acceptable to have vxworks7 support. 

Another way to look at it -- the same problem (probably) currently exists and can be reproduced in VxWorks 6.9 if you add `-Wconversion` to the command line.  Or even on native/POSIX if you add this flag.

So the warnings aren't really a VxWorks 6/7 issue at all - its a compiler settings issue.  The fact that VxWorks 7 uses a different compiler with different defaults just muddies that up, because we didn't set it explicitly.

klystron78:
Pushed without the Wconv stuff.

jphickey:
FYI - I went ahead and wrote a separate issue for the `-Wconversion` stuff -- see #660.  A future PR for that would be associated with that issue number but I recommend waiting until we discuss in CCB and decide which way to go.

jphickey (commented):
Looks good to me now, at least from inspection point of view.  

Still need to make sure coverage test is working and covering what it is supposed to (not clear if the directory implementation might need a coverage update due to use of the POSIX version now).

skliper:
With the addition of vxworks7 support to the open source repo, we need a way to confirm it works in CI or at minimum a manual setup/run (that hopefully could be implemented in a runner or similar).  Is there a way to build and run the OSAL functional tests using the vxworks SDK?  If this isn't possible, at the very minimum we need a setup available locally to build/test (what we do for vxworks 6.9).

astrogeco:
[CCB 20201202](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-December-02-2020) will revisit next week
- need some updates to coverage check based on comments

jphickey (changes requested):
OK!  The code changes look good now.

Only remaining issue is that this needs a squash merge up to the current "main" branch.  I ran into a build/dependency issue when I tried to build this locally.

Recommendation is to fetch the latest main, create/checkout a new branch on main, then do `git merge --squash vxworks7-2` ... fix conflicts and build/confirm  ... then `git push $fork $new_branch:vxworks7-2` to update this PR.

jphickey:
Also when writing a new commit message for the squashed branch - please use the basic commit message template with:

    Fix #610,  <short description>

    <long description>

This should be the pattern used for all commits.

jphickey (approved):
Looks good!

astrogeco:
- We need to test this with VxWorks 6
- Need test setup for VxWorks 7

astrogeco:
@jphickey can you help me with this conflict? feel free to merge into integration-candidate when fixed

astrogeco:
> @jphickey can you help me with this conflict? feel free to merge into integration-candidate when fixed

I think I fixed them look at line 237 in `src/os/vxworks/src/os-impl-tasks.c` in https://github.com/nasa/osal/pull/690

PR #617, opened 2020-10-02
Fix #556, Deprecate OS_open and OS_creat

### LINKED ISSUE #556: Deprecate OS_open and OS_creat
For historical/backward compatibility reasons, the API of these two functions doesn't follow the typical flow.  Rather than providing a `uint32` ID output buffer as the first argument with a separate int32 return code, they return the OSAL ID cast as an `int32` on success.  For these functions, the caller is expected to check if the result is negative, and if so, consider it an error code.  Whereas if it is non-negative, the caller is expected to cast it back to a `uint32` type and interpret it as an OSAL ID.

These should be like all other OSAL APIs and pass back the ID separately from the return/error status.

Leave as is.  But these two functions present a challenge when making a distinct type for OSAL IDs .

**Additional context**
In the current implementation,. these are just compatibility wrappers anyway.  They both call `OS_OpenCreate()` internally, which provides both open (existing file) and creat (new file) based on the flags it was passed.  The `OS_OpenCreate` function _does_ follow the correct pattern so one option would be to just expose this to the public API.

The other option is to create a new version of OS_open and OS_creat which follow the correct pattern.  But in order to provide a transition they would have to use different names.

**Requester Info**

jphickey:
**Describe the contribution**

These functions are replaced by OS_OpenCreate, which implements both functions via flags, and follows the correct OSAL API patterns.

Fixes #556 

**Testing performed**
Build and run all tests, sanity check CFE
Check coverage of file-related APIs (still 100%)

**Expected behavior changes**
`OS_open()` and `OS_create()` are deprecated.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
This initially puts the items inside an `OSAL_OMIT_DEPRECATED` block for review and testing, like has been done traditionally.
Should discuss at CCB whether this should become a hard-cut given that the next release will be a major one.
Although the previous discussions on the topic agreed to hard cut at major releases, these are such widely-used routines that this will cause grief to a lot of users.

### DIFF
--- a/src/os/inc/osapi-os-filesys.h
+++ b/src/os/inc/osapi-os-filesys.h
@@ -178,6 +178,8 @@ typedef enum
  * @{
  */
 
+#ifndef OSAL_OMIT_DEPRECATED
+
 /*-------------------------------------------------------------------------------------*/
 /**
  * @brief Creates a file specified by path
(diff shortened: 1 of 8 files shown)


astrogeco:
[CCB 2020-10-06](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-October-06-2020) APPROVED 
- This will break many apps. We'll leave it in for now but it WILL be removed before tagging osal v6.0

PR #634, opened 2020-10-23
Fix #623, add debug messages for mutex double locks

### LINKED ISSUE #623: Add OS_DEBUG warning if task locks a mutex multiple times
As noted in nasa/CFE#948, there are cases in CFE where a mutex may be taken by the same task more than once.  This is technically allowed because OSAL enables recursive mutexes for all supported OS types.

But my concern is that:

- recursive mutexes aren't guaranteed to be provided - i.e. some minimal RTOS's might not have this facility at all. 
- if recursive mutexes are provided, they are slower than the normal (non-recursive) counterpart.
- anything that depends on recursive mutexes generally indicates deeper code design flaws - i.e. insufficient separation of internal vs. external APIs, function loops, circular subsystem dependencies, etc.

Start by introducing an `OS_DEBUG` warning if the same task takes a mutex more than once.  This can help identify the issues, and it is very easy to add.  It also won't be displayed unless OSAL is compiled with debug messages enabled.

**Additional context**
A future / follow up change, once dependencies on recursive mutexes are fix, could be to allow the user to optionally elect to use non-recursive/fast mutexes via an osconfig option.

**Requester Info**

jphickey:
**Describe the contribution**

If `OS_DEBUG` is enabled, this adds a message if mutex give/take actions occur outside the expected sequence.  In particular,
this warns if a task takes a mutex more than once.

Fixes #623 

**Testing performed**
Build and sanity test CFE with debug events enabled, confirm normal boot and operation.
Send "remove packet" request to TO_LAB.

Confirm warning is generated as follows (this is a known issue in CFE doing multiple locks):

    OS_MutSemTake():216:WARNING: Task 65547 taking mutex 327685 while owned by task 65547

**Expected behavior changes**
If debugging is enabled this now informs the user (via the debug console) if a lock is taken more than once.
It will also warn if a lock is given by a different task than what originally took it.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/src/os/shared/inc/os-shared-mutex.h
+++ b/src/os/shared/inc/os-shared-mutex.h
@@ -32,7 +32,8 @@
 
 typedef struct
 {
-    char obj_name[OS_MAX_API_NAME];
+   char      obj_name[OS_MAX_API_NAME];
+   osal_id_t last_owner;
 } OS_mutex_internal_record_t;
 
 /*
--- a/src/os/shared/src/osapi-mutex.c
+++ b/src/os/shared/src/osapi-mutex.c
@@ -162,11 +162,24 @@ int32 OS_MutSemGive(osal_id_t sem_id)
     OS_common_record_t *record;
     uint32              local_id;
     int32               return_code;
+    osal_id_t           self_task;
 
     /* Check Parameters */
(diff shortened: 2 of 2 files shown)


zanzaben (approved):

astrogeco:
[CCB 2020-10-28](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-October-28-2020) APPROVED

PR #638, opened 2020-11-02
Fix #637, fix OS_ModuleUnload for static modules

### LINKED ISSUE #637: OS_ModuleUnload() for statically loaded module may segfault
**Describe the bug**
Calling `OS_ModuleUnload()` on a statically loaded module will still attempt to call `dlclose()` but the handle is NULL because no module was loaded.  At least on Linux/Glibc this actually causes a segmentation fault.

**To Reproduce**
Using the CFE framework switch any app (e.g. sample_app) from a regular dynamic app to a static app.
Build and run CFE, then shutdown with CTRL+C.
The shutdown ends up calling `OS_DeleteAllObjects()` which in turn will trigger this bug - one gets a segfault instead of a clean exit.

**Expected behavior**
Should shutdown cleanly.

**System observed on:**
Ubuntu 20.04

**Reporter Info**

jphickey:
**Describe the contribution**
Ensure that the handle is not NULL before invoking `dlclose()`.  In particular the handle will be NULL for static modules.

Fixes #637 

**Testing performed**
Build and sanity test CFE with apps linked statically.

**Expected behavior changes**
- Shutdown after CTRL+C occurs normally (no segfault)
- Also confirmed that specifying an incorrect entry point symbol in startup script also benign.  (because a bad entry point also causes module to be "unloaded" as part of cleanup, which triggered the bug too)

**System(s) tested on**
Ubuntu 20.04 
i686-rtems4.11 / pc686 via QEMU

**Additional context**
This issue only affects statically linked apps, which "go through the motions" of module loading for consistency in operation, but do not actually load a module - because its already linked.

### DIFF
--- a/src/os/shared/inc/os-shared-module.h
+++ b/src/os/shared/inc/os-shared-module.h
@@ -30,12 +30,20 @@
 
 #include <os-shared-globaldefs.h>
 
+typedef enum
+{
+    OS_MODULE_TYPE_UNKNOWN = 0, /**< Default/unspecified (reserved value) */
+    OS_MODULE_TYPE_DYNAMIC = 1, /**< Module is dynamically loaded via the OS loader */
+    OS_MODULE_TYPE_STATIC  = 2  /**< Module is statically linked and is a placeholder */
+} OS_module_type_t;
+
 typedef struct
 {
-    char    module_name[OS_MAX_API_NAME];
-    char    file_name[OS_MAX_PATH_LEN];
-    uint32  flags;
-    cpuaddr entry_point;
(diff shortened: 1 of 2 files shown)


astrogeco:
[CCB 2020-11-04](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-November-04-2020) APPROVED
- See documentation in cFE/cmake/sample_defs for doing a static build
  - Need to predeclare which symbols OS_Symbol_lookup will look for in sample_defs/targets.cmake

acudmore (approved):
This looks good to me. I like the support for static modules.

PR #643, opened 2020-11-04
Fix #641, add OS_ModuleLoad flags

### LINKED ISSUE #641: Add OS_ModuleLoad flags to indicate symbol table visibility
On POSIX platforms with a real dynamic loader/shared object implementation, using the `RTLD_GLOBAL` flag to dlopen() can make it tricky or impossible to unload modules later.  This flag makes the symbols globally available to satisfy other relocations, and unloading of the module will be deferred or prevented entirely as long as the runtime loader thinks a symbol is being used.

This ultimately causes a requirement failure on this platform as documented in nasa/cfe#952 - because the module isn't _actually_ unloaded when `dlclose()` is called, and even though the new/replacement app module was loaded, CFE will end up restarting the original code, not the new (reloaded) code.

Loading a module with `RTLD_LOCAL` instead seems to prevent this issue - because the symbols are simply not made available for other modules/entities to use.  This ensures that when the time comes to unload the module, nothing else is referencing the module, and `dlclose()` actually does unload it.

But this LOCAL flag cannot be used for all modules, because libraries do need their symbols added to the global table, or else it will not be possible to load apps that depend on those libraries 

Add a "flags" parameter onto the existing `OS_ModuleLoad()` API, so it becomes:

    int32 OS_ModuleLoad(osal_id_t *module_id, const char *module_name, const char *filename, uint32 flags)

The "flags" parameter can be used to indicate the symbol visibility.  A flag value of 0 should map to "global" - which is what the current implementation does - to make an easy transition for existing code.

In order to be able to look up an entry point in a module loaded with this option, this necessitates another new API:

    int32 OS_ModuleSymbolLookup(osal_id_t module_id, cpuaddr *SymbolAddress, const char *SymbolName)

Which is the same as `OS_SymbolLookup()` but accepts a module ID value and operates on that module, rather than on the global scope.  This should be the ID that was returned from the OS_ModuleLoad call.

For RTOS implementations that do not have this symbol visibility option they can ignore the flag, continue to map everything into the global symbol table as they currently do, and `OS_ModuleSymbolLookup` and `OS_SymbolLookup` become equivalent.

**Additional context**
Note that most all other OSAL "create" functions (tasks, queues, semaphores, etc) already have a "flags" parameter on the API, reserved for future use.  Unfortunately, this flags parameter was not part of the original `OS_ModuleLoad()` API definition, making it an exception to the pattern.  So by adding this, although it is a breaking change, it makes it more consistent with the rest of the APIs.

The alternative would be to define a separate `OS_ModuleLoadWithFlags()` API, but this pattern does not exist anywhere else, so it would continue to be an exception with respect to the overall OSAL API.

**Requester Info**

jphickey:
**Describe the contribution**
Add a "flags" parameter to OS_ModuleLoad() and use this to indicate the desired symbol visibility - either GLOBAL (0, the default, and matches current behavior) or LOCAL, which means symbols are hidden from other modules.

This is intended as a "hint" to OSAL as to how the module will be used - LOCAL prevents other modules from binding to symbols in this module, thereby ensuring/preserving the ability to unload in the future.  CFE should use LOCAL flag for apps, and GLOBAL flags for libraries, which are typically not ever unloaded.

Fixes #641

**Testing performed**
Build and run unit test
Sanity check CFE (update/patch required update to CFE to add flags parameter on API call)

**Expected behavior changes**
No impact to behavior if flags are passed as 0.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
Related to nasa/cfe#952

### DIFF
--- a/src/os/inc/osapi-os-loader.h
+++ b/src/os/inc/osapi-os-loader.h
@@ -34,6 +34,44 @@
 ** Defines
 */
 
+/**
+ * @brief Requests OS_ModuleLoad() to add the symbols to the global symbol table
+ *
+ * When supplied as the "flags" argument to OS_ModuleLoad(), this indicates
+ * that the symbols in the loaded module should be added to the global symbol
+ * table.  This will make symbols in this library available for use when
+ * resolving symbols in future module loads.
+ *
+ * This is the default mode of operation for OS_ModuleLoad().
+ *
+ * @note On some operating systems, use of this option may make it difficult
+ * to unload the module in the future, if the symbols are in use by other entities.
+ *
+ */
+#define OS_MODULE_FLAG_GLOBAL_SYMBOLS       0x00
+
+/**
+ * @brief Requests OS_ModuleLoad() to keep the symbols local/private to this module
+ *
+ * When supplied as the "flags" argument to OS_ModuleLoad(), this indicates
+ * that the symbols in the loaded module should NOT be added to the global
+ * symbol table.  This means the symbols in the loaded library will not available
+ * to for use by other modules.
+ *
+ * Use this option is recommended for cases where no other entities will need
+ * to reference symbols within this module.  This helps ensure that the module
+ * can be more safely unloaded in the future, by preventing other modules from
+ * binding to it.  It also helps reduce the likelihood of symbol name conflicts
+ * among modules.
+ *
+ * @note To look up symbols within a module loaded with this flag, use
+ * OS_SymbolLookupInModule() instead of OS_SymbolLookup().  Also note that
+ * references obtained using this method are not tracked by the OS; the
+ * application must ensure that all references obtained in this manner have
+ * been cleaned up/released before unloading the module.
+ */
+#define OS_MODULE_FLAG_LOCAL_SYMBOLS        0x01
+
 /*
 ** Typedefs
 */
@@ -107,6 +145,25 @@ int32 OS_SymbolLookup(cpuaddr *symbol_address, const char *symbol_name);
 
 /*-------------------------------------------------------------------------------------*/
 /**
+ * @brief Find the Address of a Symbol within a module
+ *
+ * This is similar to OS_SymbolLookup() but for a specific module ID.
+ * This should be used to look up a symbol in a module that has been
+ * loaded with the #OS_MODULE_FLAG_LOCAL_SYMBOLS flag.
+ *
+ * @param[in]  module_id      Module ID that should contain the symbol
+ * @param[out] symbol_address Set to the address of the symbol
+ * @param[in]  symbol_name    Name of the symbol to look up
+ *
+ * @return Execution status, see @ref OSReturnCodes
+ * @retval #OS_SUCCESS @copybrief OS_SUCCESS
+ * @retval #OS_ERROR if the symbol could not be found
+ * @retval #OS_INVALID_POINTER if one of the pointers passed in are NULL
+ */
+int32 OS_ModuleSymbolLookup(osal_id_t module_id, cpuaddr *SymbolAddress, const char *SymbolName);
+
+/*-------------------------------------------------------------------------------------*/
+/**
  * @brief Dumps the system symbol table to a file
  *
  * Dumps the system symbol table to the specified filename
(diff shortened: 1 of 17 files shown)


jphickey:
This can be reviewed if there is time.... mainly looking for feedback regarding addition of the flags parameter.

This changeset to OSAL is fully functional (contains all unit test updates, all passing, etc) but if accepted it will need a corresponding patch to CFE to pass in a value for flags.

astrogeco:
[CCB 2020-11-04](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-November-04-2020) APPROVED
- BREAKING CHANGE, 
  - though there shouldn't be many apps using this.
  - this will also break apps that are loading things incorrectly
- VxWorks doesn't have a way to do the local module symbol table lookup because of the flat namespace
- PIPE DREAM: If we run cFS and apps within a single RTP then it would be possible to do the local module lookup in VxWorks

acudmore (approved):
Looks good.. Nothing needed in the RTEMS implementation ?

jphickey:
> Looks good.. Nothing needed in the RTEMS implementation ?

Correct!  AFAIK RTEMS has a flat namespace, like VxWorks - so no change is needed for its "dlopen()" call.  It will ignore the flag.
For symbol lookups it uses the `os-impl-posix-dl-symtab.c` implementation so it is covered that way.

I built and sanity checked my RTEMS 4.11.3, seems to boot and run fine, so long as you've also merged nasa/cfe#1000.

skliper (commented):
Is there a functional test?  If not, should add to go with the new APIs.  Could be done as a separate issue.

astrogeco:
> Is there a functional test? If not, should add to go with the new APIs. Could be done as a separate issue.

@jphickey will you add these to this PR? If not can you open a new PR for them?

jphickey:
> > Is there a functional test? If not, should add to go with the new APIs. Could be done as a separate issue.
> 
> @jphickey will you add these to this PR? If not can you open a new PR for them?

I don't think we need to add anything more?  The current functional tests (in `unit-tests`) was updated to invoke the function with flags set both ways, so it is part of the functional test.  This also calls the new function (OS_ModuleSymbolLookup) - so its all in there.

But we cannot verify symbol visibility because it won't necessarily be consistent across platforms (i.e. the underlying OS may or may not actually implement the capability).  So as a result the only thing we can actually _vaildate_ in functional test remains the current operation.

astrogeco:
> But we cannot verify symbol visibility because it won't necessarily be consistent across platforms (i.e. the underlying OS may or may not actually implement the capability). So as a result the only thing we can actually _vaildate_ in functional test remains the current operation.

Sounds good to me, I'll merge this then. @skliper do we need to document this testing limitation anywhere else other than in this thread?

skliper:
The functional test should report if the underlying OS supports the functionality or not (basically check the result, if it works report as such and if it doesn't report that it's not supported).  It's not a "Failure" for it not to be supported, but should report.

skliper:
Could also make it one of the tests that require analysis to fully pass... worth a check in the end to confirm it's the expected result.  It would be an issue if it's expected to work and doesn't.

jphickey:
> The functional test should report if the underlying OS supports the functionality or not (basically check the result, if it works report as such and if it doesn't report that it's not supported). It's not a "Failure" for it not to be supported, but should report.

This flag is really just a hint to the OSAL about how the app plans to use the module being loaded.  The OSAL may or may not actually do anything differently.  Since functional tests only use public APIs, there is no way for a functional tests to determine if its supposed to have an effect or not.

The expected pattern is:

- When using `OS_ModuleLoad()` flags=GLOBAL -> app then looks up symbols with  `OS_SymbolLookup()` 
- When using `OS_ModuleLoad()` flags=LOCAL -> app then looks up symbols with `OS_ModuleSymbolLookup()`

And this is the pattern used/implemented by the functional tests now, and as such it is confirmed to work correctly.

But the issue is that an app doing this "wrongly" i.e. using LOCAL load flags with `OS_SymbolLookup()` or GLOBAL load flags with `OS_ModuleSymbolLookup()` is not guaranteed to produce an OSAL error that can be checked for in a test.  It might trigger an error, if the OS has a local symbol namespace and can do this.  But for VxWorks and RTEMS it will likely still work OK.  That would be a bug in the app though, not in the OSAL.

So while we could add an OSAL test to call the "wrong" combinations of these functions - we'd have to pass the test either way - or at least it will be an MIR result regardless of what OSAL returned.  So I don't see much value in doing this.

skliper:
I misinterpreted "But we cannot verify symbol visibility..." to mean it isn't being checked functionally for the correct behavior.  As long as the functional test is confirming the correct behavior for the correct combination I'm good with it (no need for testing incorrect combination).

jphickey:
> I misinterpreted "But we cannot verify symbol visibility..." to mean it isn't being checked functionally for the correct behavior. As long as the functional test is confirming the correct behavior for the correct combination I'm good with it (no need for testing incorrect combination).

Right - by that statement I only meant that if a module is loaded w/LOCAL - we cannot verify that its symbols are _not_ also visible via the global lookup function.  They might be, and its not a failure if they are.

PR #646, opened 2020-11-06
Fix #559, Rename set force fail

### LINKED ISSUE #559: UT_SetForceFail is a misnomer
**Describe the bug**
UT_SetForceFail assumes that the value given, that is being set as return value, is a fail condition.  There may be sometimes that a function returns more than 1 value that is not considered a fail.

**To Reproduce**
Use in a unit test where there is a stubbed function with more than 1 "successful" return value.

**Expected behavior**
Change to something like UT_SetForcedReturnValue to make it more generic.  

**Code snips**
N/A

**System observed on:**
RHEL 7.6

**Additional context**
Not a required or debilitating situation, but a name change may be make the function's effects more clear.

**Reporter Info**
Alan Gibson NASA-GSFC/587

zanzaben:
**Describe the contribution**
Fixes #559 by changing UT_SetForceFail to UT_SetDefaultReturnValue

**Testing performed**
Build and run unit test

**Expected behavior changes**
No impact to behavior

**System(s) tested on**
 Ubuntu 20.04

**Additional context**

Alex Campbell - NASA/GSFC


skliper (approved):
Recommending a ticket to change UT_ClearForceFail to UT_ClearDefaultReturnValue.

jphickey:
I'd also recommend instead of using a simple `#define` for backward compatibility - make it an inline function e.g.:

```
void UT_SetForceFail(UT_EntryKey_t FuncKey, int32 Value)
{
    UT_SetDefaultReturnValue(FuncKey, Value);
}
```

If nothing else this just makes it more like the patterns used elsewhere.  The whole thing can go into a `#ifndef OSAL_OMIT_DEPRECATED` block.

skliper:
I recommend avoiding merging main into a feature branch.  Preference is to rebase the feature on main.  You can always do a soft reset and recommit to fix this.

PR #652, opened 2020-11-10
Integration Candidate: 2020-11-10

### LINKED ISSUE #637: OS_ModuleUnload() for statically loaded module may segfault

### LINKED ISSUE #641: Add OS_ModuleLoad flags to indicate symbol table visibility

astrogeco:
**Describe the contribution**

Fix #637, static module unload fix
Fix #641, add flags to OS_ModuleLoad

**Testing performed**
Bundle CI on astrogeco: https://travis-ci.com/github/astrogeco/cFS/builds/200387877

**Expected behavior changes**

PR #638 -  Ensure that the handle is not NULL before invoking dlclose(). In particular the handle will be NULL for static modules. Shutdown after CTRL+C occurs normally (no segfault).

PR #643 - Add a "flags" parameter to OS_ModuleLoad() to indicate the desired symbol visibility:
- GLOBAL (0, the default, and matches current behavior)
- LOCAL which hides from other modules and prevents other modules from binding to symbols in this module, thereby ensuring/preserving the ability to unload in the future

CFE should use LOCAL flag for apps, and GLOBAL flags for libraries.

**System(s) tested on**

**Additional context**
Part of <https://github.com/nasa/cFS/pull/156>

**Third party code**
None

@jphickey


astrogeco:
@jphickey Got a test failure 
https://travis-ci.com/github/astrogeco/cFS/jobs/432694405
```
/home/travis/build/astrogeco/cFS/cfe/fsw/cfe-core/src/es/cfe_es_apps.c: In function ‘CFE_ES_LoadModule’:

/home/travis/build/astrogeco/cFS/cfe/fsw/cfe-core/src/es/cfe_es_apps.c:371:22: error: too few arguments to function ‘OS_ModuleLoad’

         StatusCode = OS_ModuleLoad ( &ModuleId,

                      ^~~~~~~~~~~~~

In file included from /home/travis/build/astrogeco/cFS/osal/src/os/inc/osapi.h:115:0,

                 from /home/travis/build/astrogeco/cFS/cfe/fsw/cfe-core/src/inc/cfe.h:44,

                 from /home/travis/build/astrogeco/cFS/cfe/fsw/cfe-core/src/inc/private/cfe_private.h:32,

                 from /home/travis/build/astrogeco/cFS/cfe/fsw/cfe-core/src/es/cfe_es_apps.c:39:

/home/travis/build/astrogeco/cFS/osal/src/os/inc/osapi-os-loader.h:198:7: note: declared here

 int32 OS_ModuleLoad(osal_id_t *module_id, const char *module_name, const char *filename, uint32 flags);

       ^~~~~~~~~~~~~

cfe-core/CMakeFiles/cfe-core.dir/build.make:75: recipe for target 'cfe-core/CMakeFiles/cfe-core.dir/src/es/cfe_es_apps.c.o' failed

make[7]: *** [cfe-core/CMakeFiles/cfe-core.dir/src/es/cfe_es_apps.c.o] Error 1

CMakeFiles/Makefile2:6951: recipe for target 'cfe-core/CMakeFiles/cfe-core.dir/all' failed

make[6]: *** [cfe-core/CMakeFiles/cfe-core.dir/all] Error 2

Makefile:140: recipe for target 'all' failed

make[5]: *** [all] Error 2

CMakeFiles/native_default_cpu1-all.dir/build.make:57: recipe for target 'CMakeFiles/native_default_cpu1-all' failed

make[4]: *** [CMakeFiles/native_default_cpu1-all] Error 2

CMakeFiles/Makefile2:361: recipe for target 'CMakeFiles/native_default_cpu1-all.dir/all' failed

make[3]: *** [CMakeFiles/native_default_cpu1-all.dir/all] Error 2

CMakeFiles/Makefile2:208: recipe for target 'CMakeFiles/mission-all.dir/rule' failed

make[2]: *** [CMakeFiles/mission-all.dir/rule] Error 2

Makefile:227: recipe for target 'mission-all' failed

make[1]: *** [mission-all] Error 2

Makefile:120: recipe for target 'all' failed

make: *** [all] Error 2

The command "make" exited with 2.
```

jphickey:
@astrogeco make sure the dependency is also pulled into the build:  nasa/cfe#1000

astrogeco:
That's what I get for not reading the comments before merging...

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

astrogeco:
@jphickey doxygen error up next

```
You must fix doxygen warnings for "usersguide" before submitting a pull request

/home/travis/build/astrogeco/cFS/osal/src/os/inc/osapi-os-loader.h:198: warning: The following parameters of OS_ModuleLoad(osal_id_t *module_id, const char *module_name, const char *filename, uint32 flags) are not documented:

  parameter 'flags'

/home/travis/build/astrogeco/cFS/osal/src/os/inc/osapi-os-loader.h:148: warning: argument 'symbol_address' of command @param is not found in the argument list of OS_ModuleSymbolLookup(osal_id_t module_id, cpuaddr *SymbolAddress, const char *SymbolName)

/home/travis/build/astrogeco/cFS/osal/src/os/inc/osapi-os-loader.h:148: warning: argument 'symbol_name' of command @param is not found in the argument list of OS_ModuleSymbolLookup(osal_id_t module_id, cpuaddr *SymbolAddress, const char *SymbolName)

/home/travis/build/astrogec

jphickey:
Doxygen issue is fixed on commit 3f5e8f3

(the force earlier was to just to revert an incomplete fix that I pushed too soon - its all good now).

PR #658, opened 2020-11-17
Fix #651, use unsigned types in VxWorks stack calculation

### LINKED ISSUE #651: VxWorks stack address calculation should use unsigned type
The VxWorks task create implementation calculates a stack base address which involves adjustment such as rounding and accounting for whether the stack grows up or down (per VxWorks requirements of `taskInit()`.

This does the calculation as integers, and currently uses the `long` type.

The risk is that if the address happens to lie in the negative range of this type, then the rounding/adjustments may need to go in the opposite direction.

The address calculations should use `unsigned long` instead of `long` just to ensure that all rounding and base address adjustments behave the same way in the event that the addresses lie in the upper half of memory (i.e. start with a 1 bit) which would put it in the negative range of a `long` type.

**Additional context**
This is really just a suspicion of a possible issue - can't really confirm/test because we don't have direct control of where these stack buffers get allocated in memory.

But either way using `unsigned` would be more correct anyway - and would simply avoid the possibility that the operation works differently depending on its value.

It probably used `long` in the first place only because that is what the arguments to `taskInit()` are declared as.

**Requester Info**

jphickey:
**Describe the contribution**
This should avoid any inconsistencies in the event that the memory address translates to a negative numeric value.

Fixes #651 

**Testing performed**
Build and sanity check CFE
Run all unit tests

**Expected behavior changes**
None - just avoids possible sign conversions.

**System(s) tested on**
Ubuntu 20.04
VxWorks 6.9

### DIFF
--- a/src/os/vxworks/inc/os-impl-tasks.h
+++ b/src/os/vxworks/inc/os-impl-tasks.h
@@ -37,7 +37,7 @@ typedef struct
     WIND_TCB tcb; /* Must be first */
     TASK_ID  vxid;
     void *   heap_block; /* set non-null if the stack was obtained with malloc() */
-    long     heap_block_size;
+    size_t   heap_block_size;
 } OS_impl_task_internal_record_t;
 
 /* Tables where the OS object information is stored */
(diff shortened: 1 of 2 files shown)


astrogeco:
[CCB 2020-11-18](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-November-18-2020) APPROVED

PR #661, opened 2020-11-25
Fix #608 - Add RTEMS 5.x support

### LINKED ISSUE #608: Add support for RTEMS 5.1 - RTEMS OSAL and pc-rtems OSAL BSP updates
The cFS Bundle currently supports RTEMS 4.11. Now that RTEMS 5.1 has been released, I would like to update the necessary components to support RTEMS 5.1 on the pc-rtems platform. This involves minor modifications to the cfe repository, the PSP repository, and the OSAL repository. These changes can be done in such a way that preserves the current RTEMS 4.11 support and adds RTEMS 5.1 support.

I would like to add support for RTEMS 5.1 while maintaining compatibility with the existing RTEMS 4.11 support. For the OSAL repository there are a few minor changes needed to allow support of RTEMS 5.1 and RTEMS 4.11:
- Minor changes in the OSAL pc-rtems BSP
- Minor changes in the RTEMS OSAL implementation files

Alternatives include:
- Not supporting newer versions of RTEMS, but several projects will depend on RTEMS 5 support.
- Dropping RTEMS 4.11 support and just making the changes needed for RTEMS 5, but there may be projects that depend on RTEMS 4.11. We can consider dropping RTEMS 4.11 support on a future release.

**Additional context**

**Requester Info**
Alan Cudmore / NASA GSFC Code 582

acudmore:
**Describe the contribution**
Fix #608

This change adds support for RTEMS 5.1 in the OSAL.
The change provides defines and necessary ifdefs so RTEMS 4.11 can continue to be supported.

**Testing performed**
Following the instructions in the pc-rtems PSP:
https://github.com/nasa/PSP/blob/main/fsw/pc-rtems/README.txt
and the updated README_RTEMS_5.txt from PSP pull request 197, I built the RTEMS toolchains, the pc686 BSPs and cFS bundle for both RTEMS 4.11 and RTEMS 5.x. I then ran the bundle on both RTEMS 5.1 and RTEMS 4.11 on QEMU x86.
Note, that for RTEMS 5, the latest RTEMS 5.x git repository has to be used, since there was an RFS file system bug that prevented the bundle from working on the pc-rtems platform.

I did not run the complete set of tests yet on RTEMS 5.1.

**Expected behavior changes**
For RTEMS 4.11 there should be no behavior changes at all.
For RTEMS 5.1 there should be no behavior changes as well.

**System(s) tested on**
 - Hardware: QEMU x86
 - OS: RTEMS
 - Versions: RTEMS 4.11 and RTEMS 5.1

**Additional context**
This change requires cFE Pull request 1031.
It is associated with PSP pull request 220, but since that contains readme files, it is not required to build.

**Third party code**
N/A

Alan Cudmore NASA/GSFC Code 582.0

### DIFF
--- a/src/bsp/pc-rtems/CMakeLists.txt
+++ b/src/bsp/pc-rtems/CMakeLists.txt
@@ -9,6 +9,13 @@ add_library(osal_pc-rtems_impl OBJECT
 	src/bsp_console.c
 )
 
+# This definition is needed for the gethostname call
+# By defining this, it avoids the need to use the -std=gnu99
+#  instead of the preferred -std=c99 GCC switch
+target_compile_definitions(osal_pc-rtems_impl PUBLIC
+   _BSD_SOURCE
+)
+
 # This BSP only works with "rtems" OS layer.
 # Confirming this reduces risk of accidental misconfiguration
 set(OSAL_EXPECTED_OSTYPE    "rtems" PARENT_SCOPE)
--- a/src/bsp/pc-rtems/src/bsp_start.c
+++ b/src/bsp/pc-rtems/src/bsp_start.c
@@ -83,10 +83,10 @@ void OS_BSP_Setup(void)
     cmdlinestr = bsp_cmdline();
 
     printf("\n\n*** RTEMS Info ***\n");
-    printf("%s", _Copyright_Notice);
-    printf("%s\n\n", _RTEMS_version);
-    printf(" Stack size=%d\n", (int)Configuration.stack_space_size);
-    printf(" Workspace size=%d\n", (int)Configuration.work_space_size);
+    printf("%s", OSAL_BSP_COPYRIGHT_NOTICE);
+    printf("%s\n\n", rtems_get_version_string());
+    printf(" Stack size=%d\n", (int)rtems_configuration_get_stack_space_size());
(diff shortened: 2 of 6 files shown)


jphickey (approved):
Good stuff! I was able to confirm the instructions and build myself an RTEMS 5 CFE system using this.  So I can confirm it basically works.

However - I am still not completely comfortable with the `_RTEMS_5_` ifdef.  Just like on the VxWorks 7 changes it would be better for future maintenance if we can avoid this.

What I would suggest is perhaps making a "compatibility" source file that is only built on RTEMS 4.11 and provides wrappers/compatibility glue for the stuff that has been renamed or modified in RTEMS5.

Will say "approve" because this does build as is.  We can discuss in CCB whether to transition to source selection here or in a follow on commit.

Note one difference from the vxworks7 stuff is that we don't have coverage tests for this.  If we did, the `_RTEMS_5_` macros would probably break (like they did on Vxworks7) .. so its not an immediate concern here, but still worthwhile to avoid, as we probably will want coverage test on RTEMS in the future.

astrogeco:
[CCB 20201202](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-December-02-2020) - APPROVED
- Some changes to copyright notice handling

skliper (approved):

Issue #647, opened 2020-11-10
Some "close" operations not doing standard delete procedure


jphickey:
**Describe the bug**
All close/delete ops should go through `OS_FinalizeDelete()` so things are cleaned up consistently.
But there remain a couple places where this is not done - `OS_close()` and its variants e.g. `OS_CloseFileByName` and `OS_CloseAllFiles`

**To Reproduce**
This is apparent if there is some action in the PSP handler for deallocation.  It will not get invoked for these close ops because it doesn't go through the common delete path.

**Expected behavior**
Should go through common path.

**System observed on:**
Ubuntu 20.04

**Reporter Info**
Joseph Hickey, Vantage Systems, Inc.

PR #666, opened 2020-12-02
Fix #648, 649, 664 - refactor all table array access

### LINKED ISSUE #648: Scrub OSAL for direct array references
In OSAL  there are quite a few places with direct array references to an index, such as:

https://github.com/nasa/osal/blob/1a826576f8dd9eb37f47da0e633b1d4712b8ac61/src/os/shared/src/osapi-sockets.c#L239

In many functions this is repeated many times over (i.e. makes several accesses into the table entry for the given item).

This should be separated to use local pointer(s) to the entry/entries in use.

First do a lookup, e.g.:

    stream = OSAL_TABLE_ENTRY(OS_stream_table, local_id);

Then use `stream->` to refer to that entry from there on, e.g.:

    if (stream->socket_type != OS_SocketType_STREAM)

**Additional context**
This makes the code a lot more readable and more maintainable.  The CFE was already scrubbed for this, so it makes sense for OSAL to also do the same.

**Requester Info**

### LINKED ISSUE #649: Incorrect call to OS_ObjectIdCompose_Impl in OS_TimerDelete
**Describe the bug**
OSAL has an incorrect setting of `first_cb` when deleting timers.

https://github.com/nasa/osal/blob/1a826576f8dd9eb37f47da0e633b1d4712b8ac61/src/os/shared/src/osapi-time.c#L432-L436

The second argument to `OS_ObjectIdCompose_Impl()` is a serial number, not a table index.  These are only the same value until table entries start to be re-used, after this they become different, and this will start to fail.  This should really not be using `OS_ObjectIdCompose_Impl()` at all here.

**To Reproduce**
- Create and delete several timebases - at least `OS_MAX_TIMEBASES` - such that table entries start to be re-used.
- Create another valid timebase for the test (do not delete).
- Create at least two timers based on this timebase
- Delete one of the timers.

At this point the ID in the timebase callback ring (`first_cb` member) may refer to an invalid entry - a timebase ID which does not exist.

**Expected behavior**
Should look up the `active_id` from the actual table entry instead - do not re-compose the ID, because `next_ref` is a table index, not a serial number.

**System observed on:**
Ubuntu 20.04

**Additional context**
It is only possible to trigger this after a rather extensive sequence of creating and deleting these resources.  So this is probably unlikely to ever occur in a real system where timers are typically created and run forever.    Should still be fixed though.

This was initially discovered by enforcing type-safety in the `osal_index_t` and `osal_id_t` - during this scrub it revealed that this was passing an `osal_index_t` to a function which is supposed to accept a serial number.  So type safety = good.

**Reporter Info**

### LINKED ISSUE #664: Incorrect parameter type in Timebase sync callback
Timebase sync callback prototype is currently defined as:

    typedef uint32 (*OS_TimerSync_t)(osal_index_t timer_id); /**< @brief Timer sync */

But indices (table position) is an internal OSAL value that shouldn't be used externally from OSAL.  More importantly, it is easy to alias, and cannot be differentiated if an object is deleted and then created again.

Use the full ID value, not the index.  So the prototype would be:

    typedef uint32 (*OS_TimerSync_t)(osal_id_t timer_id); /**< @brief Timer sync */

Leave as is.

**Additional context**
I was going to roll this into a larger change, but figured this technically qualifies as an "API change" so writing it as a separate issue for specific awareness.  However, nothing outside of OSAL itself (and the included tests) actually implements a sync callback.  PSPs could, but none currently do, so this really shouldn't have any current impact to users.  But if users do start using this option, better to have the full ID value than just the index.

**Requester Info**

jphickey:
**Describe the contribution**
Refactor the table array access across OSAL.  Multiple fixes combined into one PR as they depend on each other.

Fix #648 :  Rather than indexing arrays directly, use a token concept in combination with a macro to obtain the table entry.  All access is then done through this table pointer.  The token contains all relevant information about an object, importantly it has the "truth" as to what actual object ID is being manipulated.  This is necessary when operations such as new/delete may change the value in the table itself - the token always has the right value.

Fix #649 : Use the full object ID in the timer call back list.  This avoids complexities (and issues) associated with trying to regenerate the ID when the list changes.

Fix #664 : Update the timer sync callback prototype.  Pass the entire OSAL ID to the sync function, not just the index.  This is technically an API change (although they are both uint32, the value changes).  AFAIK this sync feature isn't actively used so this shouldn't really have an impact though.

**Testing performed**
Build and run all tests for native (POSIX), rtems 4.11 and VxWorks

**Expected behavior changes**
No impact to behavior - just a substantial cleanup of the internal patterns.

**System(s) tested on**
Ubuntu 20.04, rtems 4.11

**Additional context**
This is also necessary as a prerequisite to a couple other OSAL issues - where the original ID needs to be preserved across ops which create/delete entities.  The token provides this.


jphickey:
Rebased to latest `main` ... should be good to go now.

astrogeco:
[CCB 2020-12-02](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-December-02-2020)
- @klystron78 mentioned that it would be nice to have a "generic functions" archive of some sort
- Will help solve a lot of other challenges in a maintainable manner

skliper (approved):

jphickey:
This also fixes #647 although that wasn't in a separate commit  -- By using an iterator this is now invoking `OS_close()` to do the work in both `OS_CloseFileByName()` and `OS_CloseAllFiles()` instead of directly deleting the entry itself.  So any side effects will be consistent between all of them.

jphickey:
Found a minor issue here where it was filling the active_id field with RESERVED value ... corrected in commit ffb098d (actually didn't cause any breakage here but it did break my next PR for next week).

astrogeco:
@acudmore can you review this?

PR #667, opened 2020-12-02
Fix #120, Exit console loop on shutdown

### LINKED ISSUE #120: vxworks osapi.c utility task doesn't exit
The vxworks osapi.c has an optional task that is started in OS_API_Init():
```
/*---------------------------------------------------------------------------
 * Name:    UtilityTask
 * Purpose: If turned on, this task will print out the messages from
 *          the OS_printf buffer at a low priority. This will mean
 *          that the functions calling OS_printf will not block due to
 *          writing data to the UART
----------------------------------------------------------------------------*/
#ifdef OS_UTILITY_TASK_ON
void UtilityTask()
...
```

However, this utility task implementation has a while(TRUE) loop with no exit logic.  For a "load and forever run it" scenario this may work, but for any other scenario where an orderly shutdown is desired by the user this means the task is never killed.  One can get kernel crashes if the vxworks module is unloaded while that thread is still writing to output.  The forever loop thread also makes line coverage difficult.

We need a robust method to always end this thread and to ensure the parent thread doesn't return to OS control while this thread is still hanging out there.

skliper:
**Describe the contribution**
Fix #120 

Replaced condition on forever loops to end on shutdown. 

**Testing performed**
Built and ran unit tests for posix, built on vxworks (planning to build on RTEMs using the latest setup instructions...)

**Expected behavior changes**
- Loop will exit on shutdown

**System(s) tested on**
 - Hardware: cFS Dev server
 - OS: Ubuntu 18.04
 - Versions: Bundle main + this change

**Additional context**
What about also checking for a semaphore error on posix/rtems (or at least non-interrupt)?  Would make it consistent with vxworks.

**Third party code**
None

Jacob Hageman - NASA/GSFC

### DIFF
--- a/src/os/posix/src/os-impl-console.c
+++ b/src/os/posix/src/os-impl-console.c
@@ -34,6 +34,7 @@
 
 #include "os-shared-idmap.h"
 #include "os-shared-printf.h"
+#include "os-shared-common.h"
 
 /*
  * By default the console output is always asynchronous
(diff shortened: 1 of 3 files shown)


astrogeco:
[CCB 2020-12-02](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-December-02-2020) APPROVED

skliper:
Added #668 to cover semtake return.  @astrogeco - anything else holding this up?

astrogeco:
> Added #668 to cover semtake return. @astrogeco - anything else holding this up?

Nope, will merge shortly

astrogeco:
> > Added #668 to cover semtake return. @astrogeco - anything else holding this up?
> 
> Nope, will merge shortly

@skliper new conflicts after base change :(

skliper:
@astrogeco fixed conflicts.

PR #669, opened 2020-12-03
type mismatch

rajszym:
Variable return_code should be of type int32.


skliper:
Whoops, duplicated in #789 (issue in #788). Closing here.

PR #671, opened 2020-12-07
Fix #670, clean up task definitions in core test

### LINKED ISSUE #670: osal core test is not a good example of how to write tasks
**Describe the bug**
The `osal-core-test.c` refers to task functions that are actually defined (not just declared) in `osal-core-test.h`

Furthermore, these tasks just implement a tight loop:
https://github.com/nasa/osal/blob/fa0cffaf4610d7bc11fc867aef325f5823c32fab/src/tests/osal-core-test/osal-core-test.h#L254-L262

Multiple Problems:
1. Against coding standards to implement code in a header.
2. Same function is defined 20 times.
3. The task has no blocking statement at all, so it spins using 100% cpu after it is started.
4. On some operating systems (e.g. POSIX), unless async task cancellation is enabled this also means the task also cannot be deleted, because it will never execute a cancellation point.

Although the `osal-core-test` program passes and attempts to delete these tasks, in reality they keep running on POSIX and do not actually exit.

**To Reproduce**
View in a debugger after creating + deleting these tasks - the threads are still there.   CPU usage is also very high (e.g. 300%) for the remainder of the test because these tasks are spinning and do not exit.

**Expected behavior**
There is no need to redefine the same function. Recommend to use the `task_generic_no_exit()` function (which already exists) for all of these tasks.  This function also uses `OS_TaskDelay()` so it does not consume max CPU and also this serves as a cancellation point thus allowing the task to be deleted normally.

**System observed on:**
Ubuntu 20.04

**Additional context**
This becomes a bigger issue when #642 is done, as the tasks effectively are un-deletable.

**Reporter Info**

jphickey:
**Describe the contribution**
Move all definitions and instantiations out of the header file. Reuse the single task definition that was already defined.

Also only define 3 instances of each type, as this is all that the test actually uses (no need for 20).

Fixes #670 

**Testing performed**
Build and run the unit tests

**Expected behavior changes**
Test Tasks do not run at 100%

**System(s) tested on**
Ubuntu 20.04


skliper (approved):

astrogeco:

PR #675, opened 2020-12-09
Fix #410, breakup includes

### LINKED ISSUE #410: Break up osapi-os-core.h into more focused includes
Spawned from #404 discussion - osapi-os-core.h is a monolithic include covering a vast array of APIs, which makes changes harder conceptually separate and manage.

Break up osapi-os-core.h - could use the deprecation process to avoid breakage where osapi-os-core.h just includes all the more focused headers.

None

**Additional context**
None

**Requester Info**
Jacob Hageman - NASA/GSFC

jphickey:
**Describe the contribution**
Break up `osapi-os-*.h` files into units which correspond to the implementation units.  The old header file names still exist for compatibility.

Fix #410

**Testing performed**
Build and run CFE and unit tests.

**Expected behavior changes**
None - just header refactoring, no actual change.

**System(s) tested on**
Ubuntu 20.04
RTEMS 4.11


jphickey:

src/bsp/shared/src/bsp_default_symtab.c
```diff
@@ -29,8 +29,9 @@
  *   which will override this default.
  */
 
-#include "osapi.h"
 #include "bsp-impl.h"
 
+#include <osapi-module.h>
```
skliper, line 34: Double quotes per coding standard?  Also all other occurrences...

src/os/inc/osapi-common.h
```diff
@@ -0,0 +1,215 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/*
+ * @file osapi-common.h
+ */
+
+#ifndef INCLUDE_osapi_common_h
```
skliper, line 25: Looks like this adds another convention... can we just pick one?

astrogeco:
- remove "INCLUDE" from the include guard
- swap <> for "" in includes since they are reserved for system libraries
- remove jpfix ifdef

jphickey:
Rebased to "main" and made all requested changes.

Note I used a regex script to replace all the include guards to match the requested pattern, as well as to use quoted filenames instead of bracketed filenames in `#include` directives.  This ended up actually catching a lot of _other_ improperly formatted guards across other parts of OSAL - not just the files originally changed.  So as a result the number of files affected here has increased.  But it should accomplish the goal of making all the include guards consistent.

src/bsp/shared/src/bsp_default_symtab.c
jphickey: Fixed in rebase - now uses double quotes.

src/os/inc/osapi-common.h
jphickey: Fixed in rebase - uses all uppercase filename with no leading/trailing underscore or prefix.

skliper (approved):

skliper:
@astrogeco can we move this one forward?  It touches everything so the quicker we can get it in the less we are going to have to deal with conflicts.

src/os/inc/osapi-os-loader.h
```diff
@@ -30,207 +30,18 @@
 #ifndef _osapi_loader_
```
astrogeco, line 30: Do we want to fix this include guard even though the file is deprecated?

astrogeco: > @astrogeco can we move this one forward? It touches everything so the quicker we can get it in the less we are going to have to deal with conflicts.

I'll merge it now and we can revisit my commit in the IC as a hotfix if we want to fix include guard styles for deprecated code.

jphickey: FWIW - I had skipped updating these files because they were deprecated... will probably remove in a future commit with the rest of the deprecated stuff in this cycle.

PR #676, opened 2020-12-09
Fix #642, synchronous task delete on POSIX

### LINKED ISSUE #642: OS_TaskDelete is not synchronous on POSIX
**Describe the bug**

**To Reproduce**

**Expected behavior**

Suggestion to achieve this:

**System observed on:**
Ubuntu 20.04

**Requester Info**

jphickey:
**Describe the contribution**
Make OS_TaskDelete actually wait for the task to exit, not simply a cancellation request.

Fix #642

**Testing performed**
Build and run CFE and all unit tests
Test the CFE "reload app" command as documented in nasa/cfe#952 which depends on having OS_TaskDelete actually exit the task

**Expected behavior changes**
No impact to external API/observed behavior.

Internally, the global tables are now _unlocked_ when using an EXCLUSIVE lock type, which is used for create and delete ops.  Instead of holding the global lock, the ID is set to "RESERVED" (FFFFFFFF) which effectively prevents any other task from obtaining a reference to that object.  This way OSAL can more safely call C library functions which block, including (but not limited to) pthread_join, while allowing ops on other records to proceed.

**System(s) tested on**
Ubuntu 20.04
RTEMS 4.11

**Additional context**
Required as part of the fix for CFE requirement documented in nasa/cfe#952.

### DIFF
--- a/src/os/shared/src/osapi-idmap.c
+++ b/src/os/shared/src/osapi-idmap.c
@@ -354,74 +354,73 @@ void OS_ObjectIdTransactionCancel(OS_object_token_t *token)
  *-----------------------------------------------------------------*/
 int32 OS_ObjectIdConvertToken(OS_object_token_t *token)
 {
-    int32  return_code    = OS_ERROR;
-    uint32 exclusive_bits = 0;
-    uint32 attempts       = 0;
+    int32               return_code = OS_ERROR;
+    uint32              attempts    = 0;
+    OS_common_record_t *obj;
+    osal_id_t           expected_id;
 
-    OS_common_record_t *obj = OS_ObjectIdGlobalFromToken(token);
+    obj         = OS_ObjectIdGlobalFromToken(token);
+    expected_id = OS_ObjectIdFromToken(token);
+
+    /*
+     * Upon entry the ID must be valid
+     */
+    if (!OS_ObjectIdIsValid(expected_id))
+    {
+        return OS_ERR_INCORRECT_OBJ_STATE;
+    }
 
     while (true)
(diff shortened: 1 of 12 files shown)


jphickey:

astrogeco:

jphickey:
Rebased to main, and also revalidated that all platforms build and test passes.

NOTE: in order to pass tests this needs #671 also merged (dependency).  Without this change then OS_TaskDelete() fails.

src/os/shared/src/osapi-idmap.c
```diff
@@ -354,74 +354,73 @@ void OS_ObjectIdTransactionCancel(OS_object_token_t *token)
...
+            return_code = OS_ERR_INVALID_ID;
             break;
         }
 
         /*
          * If we get this far, it means there is contention for access to the object.
-         *  a) we want to increment refcount but an exclusive is pending
-         *  b) we want exclusive but refcount is nonzero
-         *  c) we want exclusive but another exclusive is pending
+         *  a) we want to some type of lock but the ID is currently RESERVED
```
skliper, line 422: typo

skliper (approved):

jphickey:
@astrogeco -- testing a preemptive merge on VxWorks and noted an issue with creating timers.  May want to hold off on merging this, I will have to add a patch of some type.

jphickey:
I have fixed the issues on VxWorks related to this and also broke it up into smaller commits.
Closing this PR, see PR #704 instead which has all this stuff plus some more fixes.

PR #678, opened 2020-12-09
Fix #673 and #677, update global locks

### LINKED ISSUE #673: POSIX unnecessarily setting signal masks in global lock
**Describe the bug**
As part of the global lock the POSIX implementation is setting the signal mask of the parent process.   However, the signal mask which is being set is the same mask that was already set in the parent process during the initialization.

**To Reproduce**
N/A

**Expected behavior**
The signal mask is already configured during initialization, so it should not be re-configured when locking/unlocking the global object tables.

**Code snips**
Locking: 
https://github.com/nasa/osal/blob/fa0cffaf4610d7bc11fc867aef325f5823c32fab/src/os/posix/src/os-impl-idmap.c#L98

Unlocking:
https://github.com/nasa/osal/blob/fa0cffaf4610d7bc11fc867aef325f5823c32fab/src/os/posix/src/os-impl-idmap.c#L149

**System observed on:**
Ubuntu 20.04

**Additional context**
I have investigated and I cannot find any justification for this anymore.  In older versions of OSAL the signal(s) were not blocked and therefore e.g. SIGINT, SIGALRM etc could be handled by any thread.  In this design it was likely important to make sure that a timer/interrupt doesn't get processed while internal OSAL ops are being performed.

But In the current OSAL design all of these signals are blocked initially in the main task - which all normal tasks then inherit.  So this is only resetting the same mask that was already configured.

**Reporter Info**

### LINKED ISSUE #677: Implement better way to wait for status change
If the object exists but is not in the correct state to access it, then the `OS_ObjectIdConvertToken()` will simply loop/retry the operation using `OS_TaskDelay()`.  This is not ideal and likely creates more delay than what is really needed.

https://github.com/nasa/osal/blob/fa0cffaf4610d7bc11fc867aef325f5823c32fab/src/os/shared/src/osapi-idmap.c#L367-L376

Should use a proper primitive to wait until the underlying state changes and wake up the task immediately.

Continue using polling w/task delays (still an option for OS types which don't have the proper primitive)

**Requester Info**

jphickey:
**Describe the contribution**
Reworks the POSIX global lock implementation.

- Does not change the POSIX signal mask when locking/unlocking the global.  Aside from being unnecessary, there was also a race condition in here (Fixes #673)

- Adds a condition variable to the global lock structure.  In the event that there is more than task competing for access to the same object, this allows the second task to be woken immediately instead of polling for a change (Fixes #677).

**Testing performed**
Build and run all unit tests
Build and sanity check CFE

**Expected behavior changes**
No longer changing signal masks repeatedly/unexpectedly.  (may be relevant to some BSP/driver developers)

**System(s) tested on**
Ubuntu 20.04
RTEMS 4.11

### DIFF
--- a/src/os/posix/src/os-impl-idmap.c
+++ b/src/os/posix/src/os-impl-idmap.c
@@ -38,7 +38,7 @@
 typedef struct
 {
     pthread_mutex_t mutex;
-    sigset_t        sigmask;
+    pthread_cond_t  cond;
 } POSIX_GlobalLock_t;
 
 static POSIX_GlobalLock_t OS_global_task_table_mut;
@@ -75,6 +75,15 @@ enum
     MUTEX_TABLE_SIZE = (sizeof(MUTEX_TABLE) / sizeof(MUTEX_TABLE[0]))
 };
 
+/*---------------------------------------------------------------------------------------
+ * Helper function for releasing the mutex in case the thread
+ * executing pthread_condwait() is canceled.
+ ----------------------------------------------------------------------------------------*/
+void OS_Posix_ReleaseTableMutex(void *mut)
+{
+    pthread_mutex_unlock(mut);
+}
+
 /*----------------------------------------------------------------
  *
  * Function: OS_Lock_Global_Impl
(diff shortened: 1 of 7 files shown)


jphickey:

astrogeco:
- Do we expect other platforms to support the condition variables from posix? There are "barriers" in RTEMS but they're not paired with mutex. VxWorks has binary semaphores but also suffers from the non atomic issue. These have a risk that the state can change in the midst of the call. 
- RTEMS does have a posix layer that might include the needed functionality
- Caution against having a difference platform implementation for a minor performance gain
- Concerns that the polling and delay would result in taking more time for this operation than expected
- @jphickey open issue about this in VxWorks and RTEMS

jphickey:
Rebased to main.   

Also fixed an intermittent issue I found in testing where the mutex would not be released.  There is a test case where a task gets deleted immediately after creation, and sometimes this would happen while it was still initializing and holding the table mutex.  Solution is to add a cleanup handler around the pthread_cond_wait just like bin sems use, so if the task is deleted, the mutex also gets unlocked.

astrogeco:
@jphickey I resolved a small conflict here, I'll do a squash-commit on this unless you want to rebase and force push.

jphickey:
> @jphickey I resolved a small conflict here, I'll do a squash-commit on this unless you want to rebase and force push.

Yes, I recall a small conflict with the #endif comment.

Does it really need a rebase though? Can't we just make a normal merge commit with the conflict resolved?  Do you want me to do that in the integration candidate?

I don't like squashes at this point because it makes it hard to track what branches in my fork can be deleted (i.e. there is no specific git record of the merge, and `git branch --merged` doesn't report it, and so on).

astrogeco:
> > @jphickey I resolved a small conflict here, I'll do a squash-commit on this unless you want to rebase and force push.
> 
> Yes, I recall a small conflict with the #endif comment.
> 
> Does it really need a rebase though? Can't we just make a normal merge commit with the conflict resolved? Do you want me to do that in the integration candidate?
> 
> I don't like squashes at this point because it makes it hard to track what branches in my fork can be deleted (i.e. there is no specific git record of the merge, and `git branch --merged` doesn't report it, and so on).

I suggested a rebase to avoid having f5b84dab54fb5ccf8a94c18540b5b6c8a4dc1f96 in the commit history. 

Regarding the squash, I think github gives me the option to delete the branch after merging

jphickey:
No worries about f5b84dab we can avoid this either way.

Regarding deleting branches I also need to delete them on my local machine - not just the fork.  I very much prefer _not_ to break `git branch --merged` command.  

I can easily/quickly push a merge commit to integration-candidate that fixes it.  Or alternatively if you really prefer a rebase I can do that too but I'd prefer to do it myself as it keeps my local branches in sync with my fork branches.

PR #682, opened 2020-12-09
Fix #569, Check and report sysconf error return

### LINKED ISSUE #569: sysconf can return -1 on error
in posix/src/os-impl-tasks.c sysconf() is used. Its storage is into a type size_t, but sysconf() can return -1. We should change the storage to ssize_t to accommodate this, or we should check for -1 and then store a non-negative value to POSIX_GlobalVars.PageSize. sysconf() shouldn't return -1 when passed _SC_PAGESIZE, so this may be low priority.

skliper:
**Describe the contribution**
Fix #569 - Checks return of sysconf for error and if so reports, only sets PageSize on success.

**Testing performed**
Build and run unit tests

**Expected behavior changes**
None since sysconf should never fail, but if it does provides mechanism to avoid error propagation

**System(s) tested on**
 - Hardware: cFS Dev Server
 - OS: Ubuntu 18.04
 - Versions: Bundle main + this commit

**Additional context**
None.

**Third party code**
None.

Jacob Hageman - NASA/GSFC

### DIFF
--- a/src/os/posix/src/os-impl-tasks.c
+++ b/src/os/posix/src/os-impl-tasks.c
@@ -206,6 +206,7 @@ static bool OS_Posix_GetSchedulerParams(int sched_policy, POSIX_PriorityLimits_t
 int32 OS_Posix_TaskAPI_Impl_Init(void)
 {
     int                    ret;
+    long                   ret_long;
     int                    sig;
     struct sched_param     sched_param;
     int                    sched_policy;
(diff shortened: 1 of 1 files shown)


astrogeco:
[CCB 2020-12-16](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-December-16-2020) APPROVED

PR #689, opened 2020-12-14
Fix #688, implement value check and bug report macros

### LINKED ISSUE #688: Implement macro to facilitate argument checking
OSAL and CFE contain many argument checks at the beginning of functions.  These would benefit from a macro to keep them consistent and readable in the application code, as well as offer alternative implementations/options for controlling how these are handled on an application-wide basis.

Provide assert-style macros in OSAL that can facilitate argument checking, bug checking, and error handling.

**Requester Info**

jphickey:
**Describe the contribution**
Provide a set of macros to facilitate the argument value checking typically performed by every public API.

- `BUGREPORT()` is a printf-like macro that reports an invalid/unexpected condition has been found which indicates a bug in the application (i.e. uncorrectable).
- `BUGCHECK()` provides a simplified method to confirm a condition is true.  If not true, then it invokes `BUGREPORT` and (possibly) returns an error code to the caller.  This is used for items which __must__ be true or else indicate serious bugs where execution cannot/should not continue normally.
- `ARGCHECK()` confirms a condition is true, but amy take a mitigation/corrective action rather than treat it as a bug if false.  This can be used for "normal" range checking which should always produce a soft failure/error code response or enforce a suitable minimum/maximum value.

Fixes #688 

**Testing performed**
Execute the "bin sem" tests - Note only the "bin sem" implementation has been updated to use these macros thus far.
Confirm the basic modes work:

1. Normal/default mode where `OSAL_CONFIG_BUGCHECK_DISABLE=false` and `OSAL_CONFIG_BUGCHECK_STRICT=false`.  In this case the error code is returned and the test passes normally.  No change to behavior, but new error printfs are visible when running the test programs and they pass in bad values.  **Note**: this is the historical behavior and the application keeps running.  This requires the application to actually check/handle the error code.  As these indicate bugs, It is quite likely that the application does not expect the error response and/or is already in a bad state such that it will not handle it correctly, causing a more obfuscated failure later on.
2. Efficient mode where `OSAL_CONFIG_BUGCHECK_DISABLE=true` (`OSAL_CONFIG_BUGCHECK_STRICT` is not used).   May be used by applications that have reached a high level of stability and have been confirmed never to invoke functions with outrightly bad arguments.  Bug checks are not performed at all, but other argument checks still are.
3. Strict mode where `OSAL_CONFIG_BUGCHECK_DISABLE=false` and `OSAL_CONFIG_BUGCHECK_STRICT=true`.  In this mode any BUGREPORT actions result in an `abort()` which in turn intentionally causes the application to generate core dump if possible (i.e. if `ulimit -c` permits).  The core file may then be opened via debugger to determine the cause of the failure with full context available.  The intent here is to catch the error early while the context/stack is still present, and avoid the obfuscation that is likely to occur with (1).

**Expected behavior changes**
No impact to behavior (beyond additional debug printfs) when using default mode.
Other modes change behavior as noted above.

**System(s) tested on**
Ubuntu 20.04


jphickey:
FOR REVIEW: only commit 2f83626 is relevant.  Will rebase with next baseline.

Looking for feedback as to whether this is the desired approach before updating other code to use the macros more extensively.  Currently only bin sem is done.

jphickey:
This can be CCB reviewed in its current form, with the following caveats:

1. It is based on an intermediate `integration-candidate` -- pending rebase to main whenever the next merge is finalized.
2. Unit test depends on the error checking in its current form (returning the expected error code).  All unit tests pass with default check settings, however if the check macros are customized/changed (either disabled or strict or some other mod) then much work is still needed in order to make the tests skip all the cases which depend on this error checking.

Fixing up the unit tests to support alternate arg check configs could take a bit of time (hundreds of test cases) so it may be worth doing as a separate PR....

astrogeco:
[CCB 2020-12-16](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-December-16-2020) APPROVED

- Replaces one-off argument checks scattered across OSAL and cFE
- Adds new items for debugging to `default_config` cmake file.

There are some situations where arguments aren't checked or could be optimized. There are some open issues for null checks. 
- @skliper post missing null-check issues here

astrogeco:
@jphickey did you create issue to fix unit-test?

skliper:
> @jphickey did you create issue to fix unit-test?

See #700

PR #687, opened 2020-12-14
Fix #686, Use errno in clock_getres error reporting

skliper:
**Describe the contribution**
Fix #686 - Now uses errno instead of status return  from clock_getres with strerror reporting

**Testing performed**
Built and ran unit tests on linux, which doesn't mean much since it doesn't include this code in functional or coverage testing.
Requesting retest from @ericgilligan-nasa

**Expected behavior changes**
Fixed error message

**System(s) tested on**
 - Hardware: cFS Dev Server
 - OS: Ubuntu 18.04
 - Versions: Bundle main + this commit

**Additional context**
None.

**Third party code**
None.

**Contributor Info - All information REQUIRED for consideration of pull request**
Jacob Hageman - NASA/GSFC

PR #659, opened 2020-11-17
Fix #679, Add Markdown version of OSAL Configuration Guide

ArielSAdamsNASA:
**Describe the contribution**
Pandoc was used to convert the OSAL Configuration Guide to a markdown file. During the conversion, issues were presented such as broken tables. Tables were fixed and block codes were added where appropriate. 

Fix #679

**Additional context**
Please provide feedback on whether the markdown file is acceptable in terms of formatting. 

**Contributor Info - All information REQUIRED for consideration of pull request**
Ariel Adams, ASRC Federal

PR #690, opened 2020-12-14
OSAL Integration Candidate: 2020-12-22

### LINKED ISSUE #377: OSAL Select APIs missing functional test

### LINKED ISSUE #410: Break up osapi-os-core.h into more focused includes

### LINKED ISSUE #542: VxWorks OS_CountSemCreate unit test fails
**Describe the bug**
The OS_CountSemCreate unit test, "#4 Initial-count-too-high", fails.  This is because the vxworks OSAL, as written, does not check the initial count.  Rather, it passes it directly to the OS call.

**Expected behavior**
It should pass.

**System observed on:**
 - AiTech SP0-S 
 - OS: VxWorks 6.9
 - OSAL 5.0.3-bv

**Reporter Info**
Mathew Benson
Windhover Labs

### LINKED ISSUE #569: sysconf can return -1 on error

### LINKED ISSUE #610: Add support for vxWorks 7

### LINKED ISSUE #670: osal core test is not a good example of how to write tasks

### LINKED ISSUE #673: POSIX unnecessarily setting signal masks in global lock

### LINKED ISSUE #679: Convert doc/OSAL-Configuration-guide.doc to markdown
Doc and PDF not GitHub friendly

Convert to markdown and remove doc/pdf

None

**Additional context**
None

**Requester Info**
Jacob Hageman - NASA/GSFC

### LINKED ISSUE #686: OSAL not reporting correct error string if clock_getres() call in OS_Posix_TimeBaseAPI_Impl_Init fails
https://github.com/nasa/osal/blob/9407cdf0be77521d850c587c0f749779c4656c3e/src/os/posix/src/os-impl-timebase.c#L228

From clock_getres man pages:
> clock_gettime(), clock_settime() and clock_getres() return 0 for success, or -1 for failure (in which case errno is set appropriately).

The return value of `clock_getres()` is just 0 or -1. If it returns -1 in the event of a failure, it sets `errno`, so the `OS_DEBUG` line should use `strerror(errno)`, not `strerror(status)`, otherwise it always reports:

```
OS_Posix_TimeBaseAPI_Impl_Init():222:failed in clock_getres: Unknown error -1
```

### LINKED ISSUE #688: Implement macro to facilitate argument checking

astrogeco:
**Describe the contribution**
Combines IC-2020-12-15 and 2020-12-22

Fix #377 Add "OSAL Select" functional tests
Fix #410, breakup OSAL includes
Fix #542, Avoid UT failure if SEM_VALUE_MAX >= UINT32_MAX
Fix #569, Check and report sysconf error return
Fix #610, Add vxworks 7 support
Fix #670, clean up task definitions in core test
Fix #673 #677, improve global lock on POSIX
Fix #679, Convert OSAL Config Guide to markdown
Fix #686, Use errno in clock_getres error reporting 
Fix #688, Implement macro to facilitate argument checking

**Testing performed**
See <https://github.com/nasa/cFS/pull/162/checks>

**Expected behavior changes**
 
PR #659 - Convert the OSAL Configuration Guide from docx and pdf to a markdown file. 

PR #671 - Test Tasks do not run at 100%. Move all definitions and instantiations out of the core-test header file and reuse the already-existing single task definition.

PR #675 - Break up `osapi-os-*.h` files into units that correspond to the implementation units. Kept old header file names for compatibility.

PR #678 - Reworks the POSIX global lock implementation. Does not change the POSIX signal mask when locking/unlocking the global. 
  - Fixes a race condition. 
  - Adds a condition variable to the global lock structure. improves handling of tasks competing for access to the same object. 
  - No longer changing signal masks repeatedly/unexpectedly. May be relevant to some BSP/driver developers.

PR #682 -   Checks return of sysconf for error and reports them. Only sets PageSize on success. If sysconf fails it provides a mechanism to avoid error propagation.

PR #687 -  Uses `errno` instead of status return from `clock_getres` with `strerror` reporting.

PR #599 - Adds support for VxWorks 7

**System(s) tested on**
Ubuntu 18.04

**Additional context**
Part of <https://github.com/nasa/cFS/pull/162>

**Third party code**
None

@jphickey 
@skliper 
@zanzaben 
@ArielSAdamsNASA


astrogeco:
@jphickey this is failing with OMIT_DEPRECATED, see <https://github.com/nasa/cFS/pull/162/checks?check_run_id=1552281978>. Was there a dependency I missed in #675 ?

```
Scanning dependencies of target cfe-core_es_UT
[ 77%] Building C object cfe-core/unit-test/CMakeFiles/cfe-core_es_UT.dir/es_UT.c.o
In file included from /home/runner/work/cFS/cFS/cfe/fsw/cfe-core/unit-test/es_UT.h:53:0,
                 from /home/runner/work/cFS/cFS/cfe/fsw/cfe-core/unit-test/es_UT.c:41:
/home/runner/work/cFS/cFS/osal/src/os/inc/osapi-os-loader.h:34:2: error: #error This header header is deprecated
 #error This header header is deprecated
  ^~~~~
cfe-core/unit-test/CMakeFiles/cfe-core_es_UT.dir/build.make:81: recipe for target 'cfe-core/unit-test/CMakeFiles/cfe-core_es_UT.dir/es_UT.c.o' failed
make[7]: *** [cfe-core/unit-test/CMakeFiles/cfe-core_es_UT.dir/es_UT.c.o] Error 1
CMakeFiles/Makefile2:6310: recipe for target 'cfe-core/unit-test/CMakeFiles/cfe-core_es_UT.dir/all' failed
```

astrogeco:
Merged https://github.com/nasa/cFE/pull/1043 which I think will fix this

jphickey:
> @jphickey this is failing with OMIT_DEPRECATED, see https://github.com/nasa/cFS/pull/162/checks?check_run_id=1552281978. Was there a dependency I missed in #675 ?

Confirmed -  nasa/cfe#1043 that you already found was the correct/intended fix for this dependency.

PR #695, opened 2020-12-17
Fix #229, mqueue functional test

### LINKED ISSUE #229: mqueue test program
OSAL should include a simple mqueue test program to validate that the user has the correct settings and permissions to create/open/close/delete mqueues. Often users stumble on mqueue configuration and it is more difficult to diagnose when it's wrapped in the entirety of OSAL/cFS.

zanzaben:
**Describe the contribution**
Fixes #229 
Generalized queue timeout test to also test message queue functionality 

**Testing performed**
Build and run the unit tests 

**Expected behavior changes**
Adds 2 seconds to the unit tests 

**System(s) tested on**
Ubuntu 20.04

Alex Campbell GSFC

### DIFF
--- a/src/tests/queue-test/queue-test.c
+++ b/src/tests/queue-test/queue-test.c
@@ -33,7 +33,9 @@ void QueueTimeoutSetup(void);
 void QueueTimeoutCheck(void);
 
 #define MSGQ_DEPTH 50
-#define MSGQ_SIZE  4
+#define MSGQ_SIZE  sizeof(uint32)
+#define MSGQ_TOTAL 10
+#define MSGQ_BURST  3
 
 /* Task 1 */
 #define TASK_1_STACK_SIZE 1024
@@ -65,8 +67,8 @@ void task_1(void)
 {
     int32  status;
     size_t data_size;
-    char   data_received[4] = {0};
-    char   expected[4]      = "xyz";
+    uint32   data_received ;
+    uint32   expected = 0;
 
     OS_printf("Starting task 1\n");
 
@@ -84,8 +86,10 @@ void task_1(void)
         if (status == OS_SUCCESS)
         {
             ++task_1_messages;
-            UtAssert_True(strcmp(data_received, expected) == 0, "TASK 1: data_received (%s) == expected (%s)",
+            UtAssert_True(data_received == expected, "TASK 1: data_received (%u) == expected (%u)",
                           data_received, expected);
+
+            expected++;
         }
         else if (status == OS_QUEUE_TIMEOUT)
         {
@@ -194,13 +198,12 @@ void QueueMessageSetup(void)
 {
     int32  status;
     uint32 accuracy;
+    int        i;
+    uint32 Data = 0;
     task_1_failures = 0;
     task_1_messages = 0;
     task_1_timeouts = 0;
-
-    int        i;
-    const char Data[4] = "xyz";
-
+  
     status = OS_QueueCreate(&msgq_id, "MsgQ", OSAL_BLOCKCOUNT_C(MSGQ_DEPTH), OSAL_SIZE_C(MSGQ_SIZE), 0);
     UtAssert_True(status == OS_SUCCESS, "MsgQ create Id=%lx Rc=%d", OS_ObjectIdToInteger(msgq_id), (int)status);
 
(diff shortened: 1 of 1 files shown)


skliper:
Looks good to me.

src/tests/queue-test/queue-test.c
```diff
@@ -64,8 +64,9 @@ void TimerFunction(osal_id_t timer_id)
 void task_1(void)
 {
     int32  status;
-    uint32 data_received;
     size_t data_size;
+    char   data_received[4] = {0};
+    char   expected[4]      = "xyz";
```
jphickey, line 69: I'd prefer if the data type stayed as `uint32` here.  It is more reflective of the way queues are actually used (most apps don't typically pass strings through queues - they are integer IDs or pointers more likely).

src/tests/queue-test/queue-test.c
```diff
@@ -83,7 +84,8 @@ void task_1(void)
         if (status == OS_SUCCESS)
         {
             ++task_1_messages;
-            OS_printf("TASK 1: Recieved a message on the queue\n");
+            UtAssert_True(strcmp(data_received, expected) == 0, "TASK 1: data_received (%s) == expected (%s)",
+                          data_received, expected);
```
jphickey, line 88: It appears this checks for the same value in all 10 messages.  This is better than nothing but it doesn't check/verify the FIFO nature of the queue.  I'd rather see a (const) list of 10 different values that are sent in sequence and then confirmed in sequence - which will confirm that the queue implementation does indeed deliver messages in order.

src/tests/queue-test/queue-test.c
```diff
@@ -172,9 +224,25 @@ void QueueTimeoutSetup(void)
     status = OS_TimerSet(timer_id, timer_start, timer_interval);
     UtAssert_True(status == OS_SUCCESS, "Timer 1 set Rc=%d", (int)status);
 
-    /* allow some time for task to run and accrue queue timeouts */
-    while (timer_counter < 100)
+    /* Put 10 messages onto the que with some time inbetween to not overfill the que*/
+    for (i = 0; i < 10; i++)
     {
-        OS_TaskDelay(100);
+        OS_TaskDelay(200);
```
jphickey, line 230: While I understand the goal of not wanting to overfill the queue - it also misses an important objective of confirming that the queue will actually queue/store data up to the requested depth limit.  This is part of the function that we really should test - that we can write multiple messages (e.g. at least 2) when the receiver is _not_ waiting to receive them - then when the receiver does call `OS_QueueGet()` later - it still gets them in the original order.  

In this test we should be guaranteed to be able to write up to `MSGQ_DEPTH` and not lose any data.  In fact this can be done in a single thread (write into queue up to `MSGQ_DEPTH`, then read queue and confirm all data came back in same order with nothing missing).

src/tests/queue-test/queue-test.c
```diff
@@ -127,28 +131,76 @@ void QueueTimeoutCheck(void)
...
     int32  status;
     uint32 accuracy;
-
     task_1_failures = 0;
     task_1_messages = 0;
     task_1_timeouts = 0;
 
+    int        i;
+    const char Data[4] = "xyz";
+
```
jphickey, line 203: nitpick - variable declarations should be grouped at beginning

jphickey (commented):
Certainly improved but there are more things I'd suggest checking (noted in comments)

src/tests/queue-test/queue-test.c
zanzaben: While it is not clear from the code this is testing that the que stores data. It sends a message every 200 milliseconds and the que doesn't start reading until after 1 second. That way when the que starts there are 5 messages waiting for it and then 5 more come spread out over 200 milliseconds each. I did it this way to test both messages waiting for the que to open and the que waiting for more messages to recieve. 

For the sake of clarity I could split that into two seperate test if we think that would be better.

src/tests/queue-test/queue-test.c
zanzaben: Changed it to send a sequence of 0 to 9.

src/tests/queue-test/queue-test.c
zanzaben: changed it not wait in the first few puts and added better comments to better express that some messages are waiting in the que.

astrogeco:

src/tests/queue-test/queue-test.c
astrogeco: do we want to revisit this?

astrogeco:
@zanzaben I think you have typos in the commit messages, there's also one comment from @jphickey that wasn't addressed.

src/tests/queue-test/queue-test.c
skliper: Data type was updated to be uint32, resolved.

src/tests/queue-test/queue-test.c
skliper: resolved

astrogeco:
Need to fix one more typo: commit 054720e3e0139ae4e9d84f9cd8b890f389137ab2 is referring to the wrong issue number

zanzaben:
> Need to fix one more typo: commit [054720e](https://github.com/nasa/osal/commit/054720e3e0139ae4e9d84f9cd8b890f389137ab2) is referring to the wrong issue number

That commit was in response to a comment here, not a different issue 
https://github.com/nasa/osal/pull/695#discussion_r551350296

PR #704, opened 2020-12-22
Fix #642, 645, 701, 702, 703 - OSAL global table management

### LINKED ISSUE #642: OS_TaskDelete is not synchronous on POSIX

### LINKED ISSUE #645: OS_TaskExit() on VxWorks causes task table owner to be the deleted task
**Describe the bug**
OS_TaskExit() calls OS_ObjectIdFinalizeDelete(). This function sets  the task's record->active_id to 0, which is also being used as the task ID on VxWorks. ObjectIdFinalizeDelete() subsequently calls OS_Unlock_Global, which errors out and does not clear the owner, as the task ID is now zeroed out and does not match the owner of the table. This does not occur on Linux

**To Reproduce**
On VxWorks, call CFE_ES_ExitChildTask with OSAL_CONFIG_DEBUG_PRINTF  enabled

**Expected behavior**
Error message should not be printed out

**Code snips**
https://github.com/nasa/osal/blob/1a826576f8dd9eb37f47da0e633b1d4712b8ac61/src/os/shared/src/osapi-idmap.c#L745-L777
https://github.com/nasa/osal/blob/1a826576f8dd9eb37f47da0e633b1d4712b8ac61/src/os/shared/src/osapi-idmap.c#L625-L676
https://github.com/nasa/osal/blob/1a826576f8dd9eb37f47da0e633b1d4712b8ac61/src/os/vxworks/src/os-impl-tasks.c#L385-L413
https://github.com/nasa/osal/blob/1a826576f8dd9eb37f47da0e633b1d4712b8ac61/src/os/posix/src/os-impl-tasks.c#L736-L751

**System observed on:**
 - SP0
 - OS: VxWorks 6.9
 - Versions: CFE 6.8, OSAL  5.1.0-rc1+dev16, 

**Additional context**
See attached screenshots for stack traces
image
image
**Reporter Info**
John N. Pham, Northrop Grumman

### LINKED ISSUE #701: Clean up inconsistent implementation of OS_Lock_Global_Impl/OS_Unlock_Global_Impl
Only the "shared" layer invokes these impl (low level) APIs - they are not invoked directly by applications, and the shared layer has already done its error checking.  Therefore they should never be invoked with a bad object type, and there is no recourse if the implementation fails.

- Make sure "shared" layer is doing the appropriate scrubbing of object type before calling "impl" layer.
- Remove duplicate/unnecessary error checking in "impl" layer that was already performed at "shared" layer.
- Make "impl" functions return `void` - these should never fail, and there is no feasible recovery if they do.  The only way they can fail is if they weren't initialized properly.  The OS_DEBUG message that is printed is the only recourse.

**Additional context**
Error checking should be consolidated at "shared" layer, layer whenever possible, as this keeps "impl" layer simpler, and also ensures that the same error checking is done on all platforms - so all platforms work the same.  Right now the error checking in these functions is not quite the same between POSIX/VxWorks/RTEMS.

**Requester Info**

### LINKED ISSUE #702: Replace remaining loops through all objects with iterators
Previous changes introduced the concept of an iterator in the OSAL shared layer to iterate across all objects of a particular type.  However there remains a few cases where code is still doing a "for" loop through each index.

For consistency of operation, these should be converted to iterators.

**Additional context**
Importantly the iterator handles locking of the global table.  Using the iterator ensures that this is done consistently, and provides the appropriate token object for any calls to other functions.

**Requester Info**

### LINKED ISSUE #703: Unlock globals during create/delete ops
For a number of resource create/delete operations, they may take some non-deterministic amount of time to execute in the system, and may require obtaining some type of mutex in the underlying C library too (e.g. if something invokes malloc).  This is especially true of creating and deleting tasks, which may need to malloc() memory for the stack, and need to create a new task resource in the kernel.

Because of this, OSAL should not hold its own global table mutex while the whole operation takes place, as this will also prevent _other_ unrelated ops from completing.

These operations should reserve an entry in the table but mark it as reserved, not with a normal ID (so distinct from a normal entry).  It can then unlock the global table and complete the non-deterministic parts of the operation, then re-lock when complete.

**Additional context**
This is required for #642, already have it implemented but felt it was worth splitting into a separate issue/commit for tracking purposes.

**Requester Info**

jphickey:
**Describe the contribution**
Address multiple issues with the OSAL global table management - general cleanup and bug fixes.

Fixes #702 - use iterators whenever possible
Fixes #645 - use an unlock key rather than task ID so OS_TaskExit() doesn't trigger a warning
Fixes #701 - general cleanup of lock/unlock impl and remove redundant logic
Fixes #703 - unlock global tables during create/delete 
Fixes #642 - keep threads "attached" in POSIX, so they can be joined when deleted.

**Testing performed**
Build and run all tests for all platforms
Sanity check CFE and also confirmed RELOAD/RESTART commands are working correctly.

**Expected behavior changes**
No longer triggers warning with OS_TaskExit() on VxWorks (see #645)
OS_TaskDelete() on POSIX does not return until the task has actually exited (see #642)

Most other changes are internal only and do not change behavior.

**System(s) tested on**
Ubuntu 20.04 (native)
RTEMS 4.11.3 + pc686
VxWorks 6.9 + mcp750

**Additional context**
Noted other issues when running some unit tests on VxWorks target that are not related to this change / preexisting issues.  Will file separate bugs about those.

### DIFF
--- a/src/os/posix/inc/os-impl-idmap.h
+++ b/src/os/posix/inc/os-impl-idmap.h
@@ -19,30 +19,26 @@
  */
 
 /**
- * \file     osapi-idmap-impl-stubs.c
- * \ingroup  ut-stubs
+ * \file     os-impl-idmap.h
+ * \ingroup  posix
  * \author   joseph.p.hickey@nasa.gov
  *
  */
 
-#include <stdio.h>
-#include <string.h>
-#include <stdlib.h>
-#include <stdarg.h>
+#ifndef OS_IMPL_IDMAP_H
+#define OS_IMPL_IDMAP_H
 
-#include "utstubs.h"
+#include "osconfig.h"
+#include "osapi-idmap.h"
+#include <pthread.h>
 
-#include "os-shared-idmap.h"
-
-/*
- * Table locking and unlocking for global objects can be done at the shared code
- * layer but the actual implementation is OS-specific
- */
-void OS_Lock_Global_Impl(osal_objtype_t idtype)
+typedef struct
 {
-    UT_DEFAULT_IMPL(OS_Lock_Global_Impl);
-}
-void OS_Unlock_Global_Impl(osal_objtype_t idtype)
-{
-    UT_DEFAULT_IMPL(OS_Unlock_Global_Impl);
-}
+    pthread_mutex_t mutex;
+    pthread_cond_t  cond;
+} OS_impl_objtype_lock_t;
+
+/* Tables where the lock state information is stored */
+extern OS_impl_objtype_lock_t *const OS_impl_objtype_lock_table[OS_OBJECT_TYPE_USER];
+
+#endif  /* OS_IMPL_IDMAP_H  */
--- a/src/os/posix/src/os-impl-idmap.c
+++ b/src/os/posix/src/os-impl-idmap.c
@@ -34,45 +34,36 @@
 #include <sched.h>
 
 #include "os-shared-idmap.h"
+#include "os-impl-idmap.h"
 
-typedef struct
+static OS_impl_objtype_lock_t OS_global_task_table_lock;
+static OS_impl_objtype_lock_t OS_queue_table_lock;
+static OS_impl_objtype_lock_t OS_bin_sem_table_lock;
+static OS_impl_objtype_lock_t OS_mutex_table_lock;
+static OS_impl_objtype_lock_t OS_count_sem_table_lock;
+static OS_impl_objtype_lock_t OS_stream_table_lock;
+static OS_impl_objtype_lock_t OS_dir_table_lock;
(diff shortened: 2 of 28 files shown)


jphickey:
Note this has a lot of the same content that was already reviewed in draft form on my previous PR #676, during CCB 2020-12-09, but broken out into several commits and also verified on VxWorks.  Notably I had to add the "RESERVED" lock type to allow time bases to be created correctly - as these do in fact need to obtain a token on a record while it is in this intermediate state.

skliper (approved):
I like the cleanup.  Didn't do a line by line review but approval all the general concepts.

astrogeco:

PR #706, opened 2020-12-28
Fix #697, use POSIX dir implementation on VxWorks6

jphickey:
**Describe the contribution**

Instead of keeping/maintaining a one-off implementation for VxWorks 6, use the POSIX implementation for this module.  The only incompatibility is the prototype for mkdir() which is missing the second argument.  This can be worked around with a simple compatibility macro that is only enabled for VxWorks 6.x builds.  This replaces previous PR #698.

Fixes #697.

**Testing performed**
Build and run all unit tests
Sanity check CFE

**Expected behavior changes**
No impact to behavior

**System(s) tested on**
Ubuntu 20.04 (native)
VxWorks 6.9 (mcp750)

**Additional context**
The actual posix dir implementation did not change at all - just using it on VxWorks 6.9 too.

Also translates/converts the coverage test cases from vxworks specific to the portable dir implementation, which benefits VxWorks7, RTEMS and POSIX too.

**Contributor Info - All information REQUIRED for consideration of pull request**
Joseph Hickey, Vantage Systems, Inc.

PR #709, opened 2020-12-28
Fix #580, improve FS_BASED mounts on VxWorks

### LINKED ISSUE #580: Possible OSAL bug in OS_FileSysAddFixedMap not allowing FS_BASED mapping on VxWorks
**Describe the bug**
Unable to add a arbitrary filesystem path mappings with OS_FileSysAddFixedMap

**To Reproduce**
Call OS_FileSysAddFixedMap w/ parameters &fsid,"/", "/", and it'll fail

**Expected behavior**
Entry added to OS_filesys_table

**Code snips**
 It looks like the error code from line 516 is propagated improperly to line 530 instead of just being used to set flags on 524, causing the finalization step to fail. 
https://github.com/nasa/osal/blob/9c74b1b5207b0eb40662731eb205dba3014de5ed/src/os/shared/src/osapi-filesys.c#L511-L530

**System observed on:**
 - SP0-s
 - OS: VxWorks 6.9
 - CFE 6.7.0+dev292, OSAL 5.0.0+dev247, PSP 1.4.0+dev76 (w/ additions for SP0-s)

**Additional context**
Attempting to add a mapping to / for netbooting off of FTP on SP0-s PSP. Mapping /cf/ to /cf/ used to work earlier this year. In addition, OS_FileSys_FindVirtMountPoint appears to fail to map to "/" if it's in the table (although I suppose it's not desirable to map / in production anyway, since it'd override all other mappings). 

**Reporter Info**
John N Pham, Northrop Grumman

jphickey:
**Describe the contribution**
The mount/unmount implementation was not adequately checking for and handling the `FS_BASED` (pass -through) mapping type - which should be mostly a no-op.  But to be consistent with POSIX it should also create a mount point directory if it does not already exist when using this mapping type.

Adds a documentation note to `OS_FileSysAddFixedMap()` regarding the limitation that the virtual mount point cannot be empty - so `OS_FileSysAddFixedMap(.., "/", "/")` does not work - never did.  However `OS_FileSysAddFixedMap(.., "/", "/root")` does work and allows one to open files in the root as "/root/<file>" from OSAL applications.

Fixes #580

**Testing performed**
Build and run all unit tests on native
Sanity check CFE
Confirm behavior of `OS_FileSysAddFixedMap()` + `OS_TranslatePath()` when mapping to root FS.

**Expected behavior changes**
Mount point directories do not need to be already existing when using `OS_FileSysAddFixedMap`

**System(s) tested on**
Ubuntu 20.04 (native)
VxWorks 6.9 (mcp750)

**Additional context**
Auto-creating the mount point dir is relevant to unit tests - this simplifies running the unit tests by not requiring the user to pre-create this directory.  Otherwise without this one gets unit test failures if the directory doesn't already exist.

### DIFF
--- a/src/os/vxworks/src/os-impl-filesys.c
+++ b/src/os/vxworks/src/os-impl-filesys.c
@@ -32,6 +32,7 @@
 #include "os-vxworks.h"
 
 #include "os-impl-filesys.h"
+#include "os-impl-dirs.h"
 #include "os-shared-filesys.h"
 #include "os-shared-idmap.h"
 
@@ -295,22 +296,62 @@ int32 OS_FileSysMountVolume_Impl(const OS_object_token_t *token)
     OS_filesys_internal_record_t *local;
     int32                         status;
     int                           fd;
+    struct stat                   stat_buf;
 
     local = OS_OBJECT_TABLE_GET(OS_filesys_table, *token);
 
     /*
-     * Calling open() on the physical device path
-     * mounts the device.
+     * For FS-based mounts, these are just a map to a some other
+     * directory in the filesystem.
+     *
+     * If it does exist then make sure it is actually a directory.
+     * If it does not exist then attempt to create it.
      */
-    fd = open(local->system_mountpt, O_RDONLY, 0644);
-    if (fd < 0)
+    if (local->fstype == OS_FILESYS_TYPE_FS_BASED)
     {
-        status = OS_ERROR;
+        if (stat(local->system_mountpt, &stat_buf) == 0)
+        {
+            if (S_ISDIR(stat_buf.st_mode))
+            {
+                /* mount point exists */
+                status = OS_SUCCESS;
+            }
+            else
+            {
+                OS_DEBUG("%s is not a directory\n", local->system_mountpt);
+                status = OS_FS_ERR_PATH_INVALID;
+            }
+        }
+        else
+        {
+            if (mkdir(local->system_mountpt, 0775) == 0)
+            {
+                /* directory created OK */
+                status = OS_SUCCESS;
+            }
+            else
+            {
+                OS_DEBUG("mkdir(%s): errno=%d\n", local->system_mountpt, errnoGet());
+                status = OS_FS_ERR_DRIVE_NOT_CREATED;
+            }
+        }
     }
     else
     {
-        status = OS_SUCCESS;
-        close(fd);
+        /*
+         * For all other (non-FS_BASED) filesystem types,
+         * Calling open() on the physical device path mounts the device.
+         */
+        fd = open(local->system_mountpt, O_RDONLY, 0644);
+        if (fd < 0)
+        {
+            status = OS_ERROR;
+        }
+        else
+        {
+            status = OS_SUCCESS;
+            close(fd);
+        }
     }
 
     return status;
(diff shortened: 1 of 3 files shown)


jphickey:
Ping @johnphamngc -- please review confirm if this is adequate to address the issue described in #580

skliper (approved):

src/os/vxworks/src/os-impl-filesys.c
```diff
@@ -295,22 +296,62 @@ int32 OS_FileSysMountVolume_Impl(const OS_object_token_t *token)
...
+            }
+            else
+            {
+                OS_DEBUG("%s is not a directory\n", local->system_mountpt);
+                status = OS_FS_ERR_PATH_INVALID;
+            }
+        }
+        else
+        {
+            if (mkdir(local->system_mountpt, 0775) == 0)
```
johnphamngc, line 327: VxWorks mkdir doesn't take a mode; throws a lint error. 

https://docs.windriver.com/bundle/VxWorks_Kernel_API_Reference_6.9_1/page/1190833.html#b89517da-114c-41d9-a160-575213a358c5__mkdir

jphickey: Sorry - I should have mentioned, if you want to build you'll need to also pull the fix/workaround for VxWorks 7 compatibility in #706 (mkdir was fixed to be POSIX compliant in VxWorks 7).

jphickey:
Added dependency label - if/when merging note that this also requires the compatibility fix in #706

src/os/vxworks/src/os-impl-filesys.c
johnphamngc: Hm, this commit https://github.com/nasa/osal/commit/25cd5dfccfdb38ab879dd661d4fab42cce7f3a21 seems to have broken vxWorks when attempting to load cFE - "Warning: module 0x56d75b0 holds reference to undefined symbol OS_WaitForStateChange_Impl." when trying to build and run the branch. I'll see if I can backport the fix to OSAL 5.1.0-rc1+dev16 and see if it fixes the issue.

jphickey: Yeah there was a bunch of PRs recently, I thought #706 + this one would work in isolation but maybe not... 

At any rate, with this change I was able to test/confirm that you can do:

`OS_FileSysAddFixedMap(&fs_id, "/", "/root");`

Which succeeds, then followed by: 

`OS_TranslatePath("/root/myfile.txt", myfile_path)`

Which in turn sets `myfile_path` to `//myfile.txt` 

Note - yes it translates to a double leading slash but that is technically OK on UNIX and VxWorks seems to accept this and open it fine as well...

johnphamngc: Can you confirm that the case of `OS_FileSysAddFixedMap(&fs_id, "/cf", "/cf");` works as well? Otherwise I'll get around to backporting the fix and checking it sometime tomorrow. 

I think I'll make another ticket for the NFS directory case, since it's unrelated besides both affecting OS_FileSysMountVolume_Impl  and is a change in the POSIX OSAL instead. That one is a lower priority since I can just tell our developers to not run cFS off of an NFS mounted directory.

johnphamngc: Looks like it works for that case as well. I'll wait till we rebase w/ the latest CFE to see if the NFS issue manifests again before making another ticket.

jphickey: Thanks, and yes please submit a separate ticket for the NFS issue.  In the meantime I am looking into getting one of my test targets booted using an NFS root, and I'll check if anything looks amiss on it.

astrogeco:

PR #710, opened 2020-12-28
Fix #708, chmod error handling

### LINKED ISSUE #708: More test failures related to "chmod" implementation
**Describe the bug**
The current `OS_FileChmod_Impl` on POSIX has some issues/limitations:

1. A while back this was changed from using the filename based calls to using a file descriptor, which "protects" against a theoretical issue where the file may get renamed while the operation is in progress.  However this introduced a potential for a file descriptor leak --  If the `fstat()` call fails, the `OS_ERROR` is returned, but the file descriptor is left open.
2. Any failure is reported as the generic `OS_ERROR` ... This should do a better job of translating the `errno` to a more specific error code.  In particular, not all file systems support unix-style file permissions - such as the very common FAT32/DOSFS - and in this case the `fchmod()` call is likely to return -1 with `errno` set to something like `ENOTSUP` or `ENOSYS`.
3. Also a file system can be mounted read only, which also prevents chmod() from working.

The main issue of 2+3 above is that the generic `OS_ERROR` code causes the chmod unit test to fail. 

**To Reproduce**
Run the chmod test on VxWorks using a DOSFS mounted filesystem (e.g. CF:0 on the MCP750 test platform).

**Expected behavior**
The chmod test cases should be skipped without failing the overall test in cases where the mounted file system does not support/allow permission to be changed.  

**System observed on:**
VxWorks 6.9 on MCP750

**Additional context**
So this is just a matter of translating the `errno` values for these conditions into `OS_ERR_NOT_IMPLEMENTED` instead of `OS_ERROR`, because the test is already implemented to check for and skip the rest of the test when it gets this return code.

**Reporter Info**

jphickey:
**Describe the contribution**
Implement better error translations in the OS_FileChmod_Impl() function.
Also corrects a file handle leak.

This makes it return OS_ERR_NOT_IMPLEMENTED when run on a file system that does not have permissions, which in turn causes the unit test to be skipped rather than fail.

Fixes #708 

**Testing performed**
Build and run all unit tests

**Expected behavior changes**
The chmod test is now skipped on VxWorks rather than failing.

**System(s) tested on**
Ubuntu 20.04 (native)
VxWorks 6.9 (mcp750)

### DIFF
--- a/src/os/portable/os-impl-posix-files.c
+++ b/src/os/portable/os-impl-posix-files.c
@@ -203,6 +203,7 @@ int32 OS_FileChmod_Impl(const char *local_path, uint32 access)
     mode_t      writebits;
     struct stat st;
     int         fd;
+    int32       status;
 
     /* Open file to avoid filename race potential */
     fd = open(local_path, O_RDONLY, 0);
@@ -211,9 +212,15 @@ int32 OS_FileChmod_Impl(const char *local_path, uint32 access)
         fd = open(local_path, O_WRONLY, 0);
         if (fd < 0)
         {
+            OS_DEBUG("open(%s): %s (%d)\n", local_path, strerror(errno), errno);
             return OS_ERROR;
-        }        
+        }
     }
+
+    /*
+     * NOTE: After this point, execution must proceed to the end of this routine
+     * so that the "fd" opened above can be properly closed.
+     */
 
     /*
      * In order to preserve any OTHER mode bits,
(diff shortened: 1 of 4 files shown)


skliper (approved):

zanzaben (approved):

astrogeco:

- skips the test if it doesn't make sense given the filesystem
- fixes a leak

PR #711, opened 2020-12-28
Fix #471, order of operations for delete all

### LINKED ISSUE #471: Order of operations on OS_DeleteAllObjects
OS_DeleteAllObjects() is used when shutting down the system, such as after an exception, a commanded processor restart, or CTRL+C, etc.

This simply deletes resources based on their numeric `OS_OBJECT_TYPE` value, meaning tasks (1) are first, followed by queues, bin/count, semaphores, mutexes, etc. and eventually timers (9).

A recent issue described in nasa/cfe#701 observed a potential problem with this.  As the task and semaphore are deleted, a timer could still be running.  If that timer executes during shutdown, it may interrupt the deletion, and attempt to use semaphore objects.

Normally this shouldn't be an issue because OSAL will return an error and reject the call.  However due to an underlying issue in Binary Semaphores (#470) after task cancellation, this caused deadlock.

It would be preferable to delete timers first, then tasks, then semaphores, files, and other resources.  This should be a safer ordering in general, as it will reduce the potential for resources to be used as they are being deleted.

Leave as-is.  

**Additional context**
The real fix for #470 prevents deadlock, this is just more future-proofing.

**Requester Info**

jphickey:
**Describe the contribution**
When cleaning up for shutdown, delete resources that have a task/thread first, followed by other resource types.  This helps avoid possible dependencies as running threads might be using the other resources.

Fixes #471

**Testing performed**
Build and run unit tests
Sanity check CFE, ensure shutdown works as expected

**Expected behavior changes**
No visible impact ... internally the tasks are deleted first during shutdown, which only has an impact if/when tasks are actively using other OSAL resources.

**System(s) tested on**
Ubuntu 20.04 (native)

### DIFF
--- a/src/os/shared/src/osapi-common.c
+++ b/src/os/shared/src/osapi-common.c
@@ -337,7 +337,18 @@ void OS_DeleteAllObjects(void)
     {
         ObjectCount = 0;
         ++TryCount;
+
+        /* Delete timers and tasks first, as they could be actively using other object types  */
+        OS_ForEachObjectOfType(OS_OBJECT_TYPE_OS_TIMECB, OS_OBJECT_CREATOR_ANY,
+                OS_CleanUpObject, &ObjectCount);
+        OS_ForEachObjectOfType(OS_OBJECT_TYPE_OS_TIMEBASE, OS_OBJECT_CREATOR_ANY,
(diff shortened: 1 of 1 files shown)


skliper (approved):

astrogeco:

PR #721, opened 2020-12-30
Fix #707 change sem flush to solve race condition

### LINKED ISSUE #707: Race condition in "select" test
**Describe the bug**
Running the OSAL select test, I ran into a deadlock situation where the "multi" test got stuck and never finished.

**To Reproduce**
Hit or miss... Run test repeatedly on a system with other loads (e.g. parallel builds)

**Expected behavior**
Test should complete

**Code snips**
Checking the test status/backtrace it looks like two tasks (main + "Server_Fn") are waiting on the binary sem.  In particular the Server_Fn is stuck here:

https://github.com/nasa/osal/blob/d698a4d7ddc66826a4c7c287468b5f1aa6ca372f/src/tests/select-test/select-test.c#L162

While the main task is waiting in the teardown code (the `TestSelectMultipleRead` has completed, and it has invoked `Teardown_Multi` which in turn invokes `Teardown_Single` here):

https://github.com/nasa/osal/blob/d698a4d7ddc66826a4c7c287468b5f1aa6ca372f/src/tests/select-test/select-test.c#L273

**System observed on:**
Ubuntu 20.04

**Additional context**
This is likely related to the use of `OS_BinSemFlush`.  We should probably deprecate this function, as I cannot see how this can ever be used safely without it being a race condition.  VxWorks offers it which (I think) is why OSAL also offers it, but its a fundamentally broken concept.

I can confirm that looking at the traceback in gdb, the flush_count is indeed already 1 - meaning the flush had already happened by the time the Server_Fn entered the bin sem take routine.

**Reporter Info**

zanzaben:
**Describe the contribution**
fixes #707
Removed Sem flush and changed it to sem give to prevent a rare race condition. 
Change the port numbers to be different from network test for when test are run in parallel. 

**Testing performed**
Ran the unit tests in parallel 

**Expected behavior changes**
Test shouldn't lock or fail 

**System(s) tested on**
Ubuntu 20.04

Alex Campbell GSFC

### DIFF
--- a/src/tests/select-test/select-test.c
+++ b/src/tests/select-test/select-test.c
@@ -108,7 +108,7 @@ void Setup_Server(void)
     UtAssert_True(actual == expected, "OS_SocketAddrInit() (%ld) == OS_SUCCESS", (long)actual);
 
     /* Set server port */
-    actual = OS_SocketAddrSetPort(&s_addr, 9997);
+    actual = OS_SocketAddrSetPort(&s_addr, 9994);
     UtAssert_True(actual == expected, "OS_SocketAddrSetPort() (%ld) == OS_SUCCESS", (long)actual);
 
     /* Set server address */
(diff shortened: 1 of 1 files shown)


jphickey (approved):

jphickey:
So far so good ... I pulled this into my OSAL baseline and ran the full test cycle several times and haven't seen it get stuck or fail yet.  (Would happen at least 1 in 3 runs previously).

skliper:
Update comment message to follow preferred format:
```
Fix #XXX, Short Description

Long Description
```
Right now it looks like:
```
Fix #XXX, 1st half of the changes

2nd halve of the changes
```

skliper (approved):

PR #735, opened 2021-01-04
Fix #429, OS_time_t with single tick counter

### LINKED ISSUE #429: Use a better time representation in OS_stat call

jphickey:
**Describe the contribution**

Use a single 64-bit tick counter to implement `OS_time_t`, rather than a split 32 bit seconds + 32 bit microseconds counter.
    
This benefits in several ways:

- increases the timing precision by 10x (0.1us ticks)
- increases the representable range by 400x (+/-14000 yrs)
- simplifies addition/subtraction (no carry over)
- avoids "year 2038" bug w/32-bit timestamps

Fixes #429 

**Testing performed**
Build and run all unit tests, sanity check CFE

**Expected behavior changes**
None

**System(s) tested on**
Ubuntu 20.04 (native)
RTEMS 4.11 + pc686 (qemu)

**Additional context**
This is the final step listed in issue #429, and it depends on several dependencies being merged already:

- PR nasa/cfe#1058
- PR nasa/psp#228
- PR nasa/osal#723

It is submitted as a separate PR from #723 due to the dependencies, so it doesn't necessarily all have to be merged at once.  However if CCB wants to accelerate the rollout this can be all done in a single merge cycle.

### DIFF
--- a/src/os/inc/osapi-clock.h
+++ b/src/os/inc/osapi-clock.h
@@ -44,9 +44,28 @@
  */
 typedef struct
 {
-    uint32 seconds;
-    uint32 microsecs;
+    int64 ticks;     /**< Ticks elapsed since reference point */
 } OS_time_t;
+
+
+/**
+ * @brief Multipliers/divisors to convert ticks into standardized units
+ * 
+ * Various fixed conversion factor constants used by the conversion routines
+ * 
+ * A 100ns tick time allows max intervals of about +/- 14000 years in
+ * a 64-bit signed integer value.
+ * 
+ * @note Applications should not directly use these values, but rather use
+ * conversion routines below to obtain standardized units (seconds/microseconds/etc).
+ */
+enum
+{
+    OS_TIME_TICK_RESOLUTION_NS = 100,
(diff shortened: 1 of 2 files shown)


wal004:
> This is the final step listed in issue #429, and it depends on several dependencies being merged already:

Would you be up for automatically managing dependent PRs? I built [Dpulls](https://github.com/marketplace/dpulls) to handle that - could be helpful if you often deal with dependent PRs - totally free for open source.

astrogeco:
[CCB:2021-01-13](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-January-13-2021) APPROVED

jphickey:
Rebased to current "main" and confirmed this builds OK against the current "main" branch of cfe/psp.  (i.e. dependencies resolved, everything looks OK).

Also added an explicit cast to `uint32` in the fractional getters (milliseconds/microseconds/nanoseconds) to avoid triggering a conversion warning in the event that some app is being compiled with `-Wconversion`.

This should be ready to go now.

PR #739, opened 2021-01-05
Fix #692, fix or statements

### LINKED ISSUE #692: Incorrect assertions in network-api-test
**Describe the bug**
The test is not correctly checking the return value in a few places.  For example:

https://github.com/nasa/osal/blob/9407cdf0be77521d850c587c0f749779c4656c3e/src/tests/network-api-test/network-api-test.c#L129

https://github.com/nasa/osal/blob/9407cdf0be77521d850c587c0f749779c4656c3e/src/tests/network-api-test/network-api-test.c#L133

**To Reproduce**
Make the subject code return something other than OS_SUCCESS or OS_ERR_NOT_IMPLEMENTED.  Assert statement will still be true.

**Expected behavior**
Probably should be `actual == OS_SUCCESS || actual == OS_ERR_NOT_IMPLEMENTED`

**System observed on:**
N/A - inspection

**Reporter Info**

zanzaben:
**Describe the contribution**
Fixes #692
fixed the or statement to correctly compare against both options. 

**Testing performed**
Build and run unit test

**Expected behavior changes**
will correctly cath failures when they happen

**System(s) tested on**
Ubuntu 20.04

Alex Campbell GSFC

### DIFF
--- a/src/tests/network-api-test/network-api-test.c
+++ b/src/tests/network-api-test/network-api-test.c
@@ -113,8 +113,14 @@ void TestDatagramNetworkApi_Setup(void)
 
     /* OS_SocketOpen */
     actual = OS_SocketOpen(&socket_id, OS_SocketDomain_INET6, OS_SocketType_DATAGRAM);
-    UtAssert_True(actual == OS_SUCCESS || actual ==  OS_ERR_NOT_IMPLEMENTED,
-                "OS_SocketOpen() (%ld) Passed", (long)actual);
+    if (actual == OS_ERR_NOT_IMPLEMENTED)
+    {
+        UtPrintf("INET6 not supported\n");
+    }
+    else
+    {
+        UtAssert_True(actual == OS_SUCCESS, "OS_SocketOpen() (%ld) Passed", (long)actual);
+    }
     OS_close(socket_id);
 
     expected = OS_INVALID_POINTER;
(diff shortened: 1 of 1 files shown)


skliper (approved):

skliper:
How about" Fix #692, Resolve typos in unit test assert OR statements"?

skliper (approved):

skliper (changes requested):
Looking closer I'd recommend something like:
```
if (actual == OS_ERR_NOT_IMPLEMENTED)
{
    UtPrintf("XXX API not implemented");
}
else
{
... do rest of the test
}
```

skliper (changes requested):
Needs auto-formatting applied, otherwise great improvement.

skliper:
Side note - not implemented isn't actually considered and "error", test passes either way (and it's OK for OSAL/OS's to not implement all API functionality).

zanzaben:
> Side note - not implemented isn't actually considered and "error", test passes either way (and it's OK for OSAL/OS's to not implement all API functionality).

Well then it shouldn't be called os_ERR_not_implemented

skliper (approved):
Good point!  In the context of FSW it is an error, and brings up that we should probably be flagging the not implemented cases as `UTASSERT_CASETYPE_NA` so they get explicitly counted/reported.  Lets address that as a separate issue though.

astrogeco:
@skliper opened related issue to make finding "not implemented" items

PR #734, opened 2021-01-04
Fix #644, Remove alignment macros

skliper:
**Describe the contribution**
Fix #644 - removes the non-portable OS_PACK and OS_ALIGNED macros

**Testing performed**
Built and ran unit tests, no issues (no dependencies in bundle)

**Expected behavior changes**
None

**System(s) tested on**
 - Hardware: cFS Dev Server
 - OS: Ubuntu 18.04
 - Versions: Bundle main + this commit

**Additional context**
None

**Third party code**
None

**Contributor Info - All information REQUIRED for consideration of pull request**
Jacob Hageman - NASA/GSFC

PR #733, opened 2021-01-04
Fix #732, change uint32 to size_t

zanzaben:
**Describe the contribution**
Fixes #732

change prototypes should be the same, from uint32 to size_t

**Testing performed**
Build on RTEMS 5

**Expected behavior changes**
Able to build

**System(s) tested on**
Ubuntu 20.04 as build host for RTEMS 5.0.0

**Additional context**
Dependant on nasa/osal#704 to successfully build

**Contributor Info - All information REQUIRED for consideration of pull request**
Alex Campbell GSFC

PR #720, opened 2020-12-29
Fix #606, Resolve cast-align error in VxWorks OS_TaskGetId_Impl

skliper:
**Describe the contribution**
Fix #606 - store taskTCB return in a `void *`, then cast to `OS_impl_task_internal_record_t *` just for the pointer math to avoid a strict alignment compiler error

Note the error was observed with cast-align=strict on Ubuntu 20.04 with ENABLE_UNIT_TESTS=true

**Testing performed**
Built without the error after fix.  Built/ran unit tests on Ubuntu 18.04 and they all passed.

**Expected behavior changes**
None other than avoid compiler error

**System(s) tested on**
 - Hardware: cFS Dev Server
 - OS: Ubuntu 18.04 (and 20.04 to confirm strict alignment error resolved)
 - Versions: Bundle + this commit

**Additional context**
None

**Third party code**
None

**Contributor Info - All information REQUIRED for consideration of pull request**
Jacob Hageman - NASA/GSFC

PR #717, opened 2020-12-29
Fix #573, add OS_FileSysStatVolume

jphickey:
**Describe the contribution**
Add OS_FileSysStatVolume as replacement for OS_fsBytesFree and OS_fsBlocksFree.

Fixes #573

**Testing performed**
Run all unit tests.

**Expected behavior changes**
New API for getting stats on file system.
Uses existing OS_FileSysStatVolume_Impl call - just exposes it in public API.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
This PR is only adding a new API, it is not (yet) deprecating the existing OS_fsBlocksFree/OS_fsBytesFree calls.  Applications and CFE will need an update before these calls can be deprecated, but the new API needs to be in place before the existing refs can be fixed.

**Contributor Info - All information REQUIRED for consideration of pull request**
Joseph Hickey, Vantage Systems, Inc.

PR #750, opened 2021-01-11
osal Integration candidate: 2021-01-12

### LINKED ISSUE #429: Use a better time representation in OS_stat call

### LINKED ISSUE #445: OS_ConvertToArrayIndex and OS_TimeBaseGetFreeRun SEGFAULT When Passed in a NULL Pointer
**Describe the bug**
OS_ConvertToArrayIndex  and OS_TimeBaseGetFreeRun SEGFAULT when passed in a NULL pointer as the second input instead of handling the NULL pointer and returning an error.

**To Reproduce**
1.  Checkout branch fix374-2-Object-Utilities-Missing-Tests in osal
2.  Go to path: cFS/osal/src/tests/idmap-api-test
3.  Build and run with the following in the main function:  
    OS_ConvertToArrayIndex(bin_sem_id, NULL);
4.  Checkout branch fix380-Add-TimeBase-Api-Functional-Tests
5.  Go to path: cFS/osal/src/tests/time-base-api-test
6.  Build and run with the following in the main function:  
    OS_TimeBaseGetFreeRun(time_base_id2, NULL);

**Expected behavior**
OS_ConvertToArrayIndex and OS_TimeBaseGetFreeRun SEGFAULT when passed in a NULL pointer, should handle the NULL pointer and return an associated error code. 

**Code snips**
(path: cFS/osal/src/tests/idmap-api-test)

    uint32 bin_sem_id;
    OS_BinSemCreate( &bin_sem_id, "BinSem", 1, 0);
   actual   = OS_ConvertToArrayIndex(bin_sem_id, NULL);
   expected = OS_ERR_INCORRECT_OBJ_TYPE;
   UtAssert_True(actual == expected , "OS_ConvertToArrayIndex() (%ld) == %ld ", (long)actual, (long)expected );

Output: 
The following tests FAILED:
	 37 - idmap-api-test (SEGFAULT)

**System observed on:**
cFS Dev Server
OS: Ubuntu 18.04
Versions: OSAL 5.0.11.0

**Reporter Info**
Yasir Khan
NASA GSFC

### LINKED ISSUE #471: Order of operations on OS_DeleteAllObjects

### LINKED ISSUE #544: DATAGRAM Network Function OS_SocketSendTo Segfault
**Describe the bug**
OS_SocketSendTo() Segfaults when passed in NULL for Remote Address (last parameter) and all the other parameters are valid.

**System observed on:**
cFS Dev Server
OS: Ubuntu 18.04
Versions: OSAL 5.0.11.0

**Reporter Info**
Yasir Majeed Khan
NASA GSFC/Emergent Space

### LINKED ISSUE #573: OS_fsBlocksFree returns a 32-bit value, but OS_statvfs_t.blocks_free is 64 bit
OS_fsBlocksFree returns a 32-bit value, but OS_statvfs_t.blocks_free is 64 bit.

### LINKED ISSUE #580: Possible OSAL bug in OS_FileSysAddFixedMap not allowing FS_BASED mapping on VxWorks

### LINKED ISSUE #606: Cast-align warning in OS_TaskGetId_Impl for vxworks
```
/home/jhageman/cFS-GitHub/osal/src/os/vxworks/src/os-impl-tasks.c: In function ‘OS_TaskGetId_Impl’:
/home/jhageman/cFS-GitHub/osal/src/os/vxworks/src/os-impl-tasks.c:411:12: warning: cast increases required alignment of target type [-Wcast-align]
  411 |     lrec = (OS_impl_task_internal_record_t *)taskTcb(taskIdSelf());
```

Recreate by building on Ubuntu 20.04, set `cast-align=strict` in `sample_defs/arch_build_custom.cmake` and prep with `ENABLE_UNIT_TESTS=true`

Resolve warnings.

None

**Additional context**
Observed on Ubuntu 20.04 w/ cast-align=strict

**Requester Info**
Jacob Hageman - NASA/GSFC

### LINKED ISSUE #642: OS_TaskDelete is not synchronous on POSIX

### LINKED ISSUE #644: Remove OS_PACK define
Shouldn't use non-standard packing options.  Really should scrub the list.

https://github.com/nasa/osal/blob/1a826576f8dd9eb37f47da0e633b1d4712b8ac61/src/os/inc/common_types.h#L52-L75

Remove/scrub

None

**Additional context**
None

**Requester Info**
Jacob Hageman - NASA/GSFC

### LINKED ISSUE #645: OS_TaskExit() on VxWorks causes task table owner to be the deleted task

### LINKED ISSUE #697: Build broken for VxWorks on RHEL 7
**Describe the bug**
The newest RHEL 7 comes by default with CMake version 2.8.12.2 in the official OS repos.
However the "VERSION_GREATER_EQUAL" comparison was added in 3.7.  Thus the build became broken on this platform after merging #610.

**To Reproduce**
Attempt to build for VxWorks using RHEL 7.x as the host platform.  Make file generation fails immediately with:

```
CMake Error at /home/jphickey/cfecfs/github/osal/src/os/vxworks/CMakeLists.txt:48 (if):
  if given arguments:

    "CMAKE_SYSTEM_VERSION" "VERSION_GREATER_EQUAL" "7.0"

  Unknown arguments specified
```

**Expected behavior**
Build should succeed.

**Code snips**
https://github.com/nasa/osal/blob/d698a4d7ddc66826a4c7c287468b5f1aa6ca372f/src/os/vxworks/CMakeLists.txt#L48

**System observed on:**
Red Hat Enterprise Linux Workstation release 7.9 (Maipo)

**Reporter Info**

### LINKED ISSUE #702: Replace remaining loops through all objects with iterators

### LINKED ISSUE #703: Unlock globals during create/delete ops

### LINKED ISSUE #708: More test failures related to "chmod" implementation

### LINKED ISSUE #732: RTEMS 5 needs size_t type to build
**Describe the bug**
There are a few methods that won't build in rtmes 5 because they use uint32 instead of size_t

Relevant Files and line numbers:
bsp_console.c 44
os-impl-network.c 60
os-impl-queues.c 159, 256
os-impl-tasks.c 376

**To Reproduce**
Build for RTEMS 5 per instructions and default config. Build fails.

**Expected behavior**
Build should succeed.

**System observed on:**
Ubuntu 20.04 as build host for RTEMS 5.0.0

**Reporter Info**
Alex Campbell GSFC

### LINKED ISSUE #755: "Select Test" still hanging
**Describe the bug**
Select test can still sometimes get stuck and hang forever. #707 fixed one cause but this is a different way it can happen. 
It's caused because it is waiting for a subtask to finish but if for whatever reason that subtask never even starts, then it gets stuck waiting forever 

**To Reproduce**
Run the test over and over until you see it happen. 

**Expected behavior**
Test should never under any circumstance hang forever. 

**Code snips**
If applicable, add references to the software.

**System observed on:**
The CI

**Reporter Info**
Alex Campbell GSFC

astrogeco:
**Describe the contribution**

Fix #702, use iterators instead of for loops
Fix #645, implement separate lock/unlock key
Fix #703, implement exclusive/reserved lock
Fix #642, make OS_TaskDelete sync
Fix #703, unit test 
Fix #697, use POSIX dir implementation on V
Fix #580, improve FS_BASED mounts on 
Fix #708, chmod error h
Fix #471, order of operations for delete all
Fix #445, add pointer parameter checks
Fix #573, add OS_FileSysSta
Fix #544, add pointer check
Fix #606, Resolve cast-align error in VxWorks OS_TaskGet
Fix #429, update OSAL code to use time ac
Fix #429, check time conversions in coverage test
Fix #644, Remove alignment macros
Fix #429, add "assemble" routines for milli/mi
Fix #732 change uint32 to size_t
Fix #755, resolve subtasks not ending on time

**Testing performed**
See <https://github.com/nasa/cFS/pull/174/checks>

**Expected behavior changes**

PR #704 
  - Address issues with OSAL global table management:
    - use iterators whenever possible
    - use an unlock key rather than task ID so OS_TaskExit() doesn't trigger a warning
    - general cleanup of lock/unlock impl and remove redundant logic
    - unlock global tables during create/delete
    - keep threads "attached" in POSIX, so they can be joined when deleted.
  - No longer triggers warning with OS_TaskExit() on VxWorks (see #645)
  - `OS_TaskDelete()` on POSIX does not return until the task has actually exited (see #642)

PR #710 
  - The chmod test is now skipped on VxWorks rather than failing. The `OS_FileChmod_Impl()` function now returns `OS_ERR_NOT_IMPLEMENTED` when run on a file system that does not have permissions, which in turn causes the unit test to be skipped rather than fail.   
  - Corrects a file handle leak. 

PR #716 
  -  Add parameter check to `OS_SocketSendTo` and adjust coverage test to validate.

PR #717 
  - Replace `OS_fsBytesFree` and `OS_fsBlocksFree` with `OS_FileSysStatVolume`. This new API for getting stats on file system. Uses existing `OS_FileSysStatVolume_Impl` call and exposes it in the public API.

PR #711 
  - When cleaning up for shutdown, delete resources that have a task/thread first, followed by other resource types. This helps avoid possible dependencies as running threads might be using the other resources. No detectable external impact; internally, the tasks are deleted first during shutdown, which only has an impact if/when tasks are actively using other OSAL resources.

PR #709
  - The mount/unmount *VxWorks* implementation was not adequately checking for and handling the `FS_BASED` pass -through mapping type - which should be mostly a no-op. Create a mount point directory if it does not already exist when using this mapping type for consistency with POSIX.
  - Adds a documentation note to `OS_FileSysAddFixedMap()`: The virtual mount point cannot be empty - so `OS_FileSysAddFixedMap(.., "/", "/")` does not work but `OS_FileSysAddFixedMap(.., "/", "/root")` does work and allows one to open files in the root as `"/root/"` from OSAL applications. Mount-point directories do not need to previously exist when using OS_FileSysAddFixedMap

PR #720 
  - store `taskTCB` return in a `void *`, then cast to `OS_impl_task_internal_record_t *` to avoid a strict alignment compiler error

PR #734 
  - Removes the non-portable `OS_PACK` and `OS_ALIGNED` macros.

PR #706 
  - Uses the POSIX dir implementation on VxWorks 6.9. The only incompatibility is the prototype for `mkdir()`which is missing the second argument; this is worked around with a compatibility macro for VxWorks 6.x builds. 
  - Translate and convert the VxWorks coverage test cases to the portable dir implementation, which benefits VxWorks7, RTEMS, and POSIX.

PR #733 
  - Fixes prototypes so they run on RTEMS by replacing uint32 with size_t

PR #715 
  - Adds` OS_CHECK_POINTER` macros to `OS_ConvertToArrayIndex` and `OS_TimeBaseGetFreeRun` so they can handle NULL pointers and return the correct error.

PR #723 
  - Adds access functions to convert/extract different units from an OS_time_t value - so that other code in CFE/PSP/Apps can be updated to use the access functions and thereby not break when the internal time definition changes. Replaces the `int32` with `OS_time_t` in the "stat" structure used by the file module. Updates the pointer argument to `OS_SetLocalTime()` to be `const`. Prototype change of `OS_SetLocalTime()` should be backward compatible.

**Additional context**
Part of <https://github.com/nasa/cFS/pull/174>

**Third party code**
None

@skliper 
@jphickey 
@zanzaben


astrogeco:
Got some doxygen errors with the last merge

```
/home/runner/work/cFS/cFS/osal/src/os/inc/osapi-filesys.h:71: warning: Unsupported xml/html tag <virt_mount_point> found
/home/runner/work/cFS/cFS/osal/src/os/inc/osapi-filesys.h:71: warning: Unsupported xml/html tag <relative_path> found
/home/runner/work/cFS/cFS/osal/src/os/inc/osapi-filesys.h:71: warning: Unsupported xml/html tag <file> found
/home/runner/work/cFS/cFS/osal/src/os/inc/osapi-filesys.h:72: warning: Unsupported xml/html tag <virt_mount_point> found
/home/runner/work/cFS/cFS/osal/src/os/inc/osapi-filesys.h:72: warning: Unsupported xml/html tag <file> found
```

astrogeco:
> @jphickey Got some doxygen errors with the last merge
> 
> ```
> /home/runner/work/cFS/cFS/osal/src/os/inc/osapi-filesys.h:71: warning: Unsupported xml/html tag <virt_mount_point> found
> /home/runner/work/cFS/cFS/osal/src/os/inc/osapi-filesys.h:71: warning: Unsupported xml/html tag <relative_path> found
> /home/runner/work/cFS/cFS/osal/src/os/inc/osapi-filesys.h:71: warning: Unsupported xml/html tag <file> found
> /home/runner/work/cFS/cFS/osal/src/os/inc/osapi-filesys.h:72: warning: Unsupported xml/html tag <virt_mount_point> found
> /home/runner/work/cFS/cFS/osal/src/os/inc/osapi-filesys.h:72: warning: Unsupported xml/html tag <file> found
> ```

Fixed, see https://github.com/nasa/osal/pull/750/commits/f9dd6b3138f5bd1999f1bf8ad109c374975d13ce

PR #766, opened 2021-01-20
Fix #742, make sure all pointers are checked for null

### LINKED ISSUE #742: Scrub API's for null pointer checks
All pointers should be checked for NULL prior to accessing, a few cases where this was observed have been fixed but need to do a complete scrub.

Scrub and add any missing null pointer checks

None

**Additional context**
Use the new macros...

**Requester Info**
Jacob Hageman - NASA/GSFC

### LINKED ISSUE #765: API's Null pointer check in void methods
Some pointers are used in methods that return void so the OS_CHECK_POINTER function doesn't work in those methods since it returns an error code. There needs to be a way to check that pointers are valid inside void methods. 

Need to discuss the best solution 

**Additional context**
related to #742, That branch has this comment marking everywhere it comes up. 
    /* TODO: void pointer, https://github.com/nasa/osal/issues/765 */

**Requester Info**
Alex Campbell GSFC

zanzaben:
**Describe the contribution**
Fixes #742 
Went through all the api's and made sure all pointers have a null check, or a comment stating that it can be null. 

Fixes #765 

**Testing performed**
Build and run unit test

**Expected behavior changes**
none

**System(s) tested on**
Ubuntu 20.04

**Additional context**
13 To Do left in here that will be fixed by #765 

Alex Campbell GSFC

### DIFF
--- a/src/os/shared/src/osapi-binsem.c
+++ b/src/os/shared/src/osapi-binsem.c
@@ -100,7 +100,7 @@ int32 OS_BinSemCreate(osal_id_t *sem_id, const char *sem_name, uint32 sem_initia
     OS_object_token_t             token;
     OS_bin_sem_internal_record_t *binsem;
 
-    /* Check inputs */
+    /* Check parameters */
     OS_CHECK_POINTER(sem_id);
     OS_CHECK_APINAME(sem_name);
 
@@ -255,7 +255,7 @@ int32 OS_BinSemGetIdByName(osal_id_t *sem_id, const char *sem_name)
 {
     int32 return_code;
 
-    /* Check inputs */
+    /* Check parameters */
     OS_CHECK_POINTER(sem_id);
     OS_CHECK_POINTER(sem_name);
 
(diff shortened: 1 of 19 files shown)


jphickey:
Note - we should __not__ be doing pointer checks on internal functions - only those that are part of the public API.

I took a glance at the proposed commit and I see it is adding checks to internal routines.  This is inefficient and unnecessary - as the value should have been already checked in the public API it was called from.  It is also my understanding that code standards only require it on externally-called functions.

skliper:
> Note - we should **not** be doing pointer checks on internal functions - only those that are part of the public API.
> 
> I took a glance at the proposed commit and I see it is adding checks to internal routines. This is inefficient and unnecessary - as the value should have been already checked in the public API it was called from. It is also my understanding that code standards only require it on externally-called functions.

Concur.  The coding standard applies to all the APIs.

zanzaben:
After removing all the internal checks this pull request is just a bunch of formatting fixes and changing comments to all use the same terminology.

skliper:
Can you add the BUGCHECKs in the void public APIs?  That would close out the issue.

zanzaben:
> Can you add the BUGCHECKs in the void public APIs? That would close out the issue.

Done

astrogeco:
[CCB:2021-01-27](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-January-27-2021) APPROVED
-

PR #810, opened 2021-02-12
Fix #808, length-limited string length checks

### LINKED ISSUE #808: Convert from strlen to strnlen where appropriate
Per security standards strlen should be avoided if possible:
```
src/os/vxworks/src/os-impl-shell.c:        OS_write(fdCmd, Cmd, strlen(Cmd));
src/os/vxworks/src/os-impl-symtab.c:    if (strlen(name) >= OS_MAX_SYM_LEN)
src/os/shared/src/osapi-sockets.c:        len = strlen(sock->stream_name);
src/os/shared/src/osapi-sockets.c:        len = strlen(sock->stream_name);
src/os/shared/src/osapi-idmap.c:            name_len = strlen(record->name_entry);
src/os/shared/src/osapi-filesys.c:    mplen = strlen(filesys->virtual_mountpt);
src/os/shared/src/osapi-filesys.c:    if (strlen(dev_name) >= OS_FS_DEV_NAME_LEN)
src/os/shared/src/osapi-filesys.c:    VirtPathLen = strlen(VirtualPath);
src/os/shared/src/osapi-filesys.c:    if (strlen(name_ptr) >= OS_MAX_FILE_NAME)
src/os/shared/src/osapi-filesys.c:            SysMountPointLen = strlen(filesys->system_mountpt);
src/os/shared/src/osapi-filesys.c:            VirtPathBegin    = strlen(filesys->virtual_mountpt);
```

Replace with strnlen where possible.

None

**Additional context**
Security analysis warning

**Requester Info**
Jacob Hageman - NASA/GSFC

jphickey:
**Describe the contribution**

Create a wrapper around `memchr()` that mimics the non-C99 function `strnlen()` which is defined in POSIX-2008.

Use this instead of `strlen()` whenever the string being checked either originates in or will be copied into a fixed-length array buffer.

Fixes #808 

**Testing performed**
Build and sanity check CFE
Run all unit tests on both native and RTEMS

**Expected behavior changes**
No behavior changes except if a bug causes strings to be unterminated.

**System(s) tested on**
Ubuntu 20.04 (native)
RTEMS 4.11.3 (qemu)

**Additional context**
Worth noting that in most cases this was testing the length of a string in the internal OSAL table entry, which was already length-checked when created.  So this check is somewhat redundant, the only way this could catch something is if there is memory corruption of some sort.

### DIFF
--- a/src/os/shared/inc/os-shared-common.h
+++ b/src/os/shared/inc/os-shared-common.h
@@ -128,4 +128,28 @@ void OS_IdleLoop_Impl(void);
  ------------------------------------------------------------------*/
 void OS_ApplicationShutdown_Impl(void);
 
+
+/*----------------------------------------------------------------
+
+   Function: OS_strnlen
+
+    Purpose: Utility function to safely find the length of a string
+             within a fixed-size array buffer.
+
+             Provides a local OSAL routine to get the functionality
+             of the (non-C99) "strnlen()" function, via the 
+             C89/C99 standard "memchr()" function instead.
+    
+ ------------------------------------------------------------------*/
+static inline size_t OS_strnlen(const char *s, size_t maxlen)
+{
+    const char *end = memchr(s, 0, maxlen);
+    if (end != NULL)
+    {
+        /* actual length of string is difference */
+        maxlen = end - s;
+    }
+    return maxlen;
+}
+
 #endif  /* OS_SHARED_COMMON_H  */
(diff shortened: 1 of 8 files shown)


jphickey:
Note - in places where it doesn't actually care what the length is and only checking it is terminated within the limit are replaced with a direct call to `memchr()` rather than going through the `OS_strnlen()` wrapper.  Places where the length is actually _used_ afterward should use the wrapper.

skliper (approved):

skliper:
@astrogeco requesting fasttrack, no behavior changes.

PR #826, opened 2021-02-19
Fix #795, Add uncovered vxworks source files to coverage statistics

### LINKED ISSUE #795: Update coverage tests for VxWorks at minimum to include reporting of all code that could be included in build
Currently os-impl-bsd-socket.c can be included (and is by default) in the VxWorks build:
https://github.com/nasa/osal/blob/09a2c5e793639e508174a2446989274acf36f25e/src/os/vxworks/CMakeLists.txt#L67

but it's not included in coverage testing or reported as part of status
https://github.com/nasa/osal/blob/09a2c5e793639e508174a2446989274acf36f25e/src/unit-test-coverage/vxworks/CMakeLists.txt#L31

So although code coverage reports 100%, we are NOT covering the possible FSW 100%... so it's a bit misleading.

All possible code should be reported for coverage.  Short term is just to add tests (track by hand) but eventually may benefit from a quick check that all the files included in the build are included in coverage analysis.

None

**Additional context**
Illuminated by investigation into #793, unreachable code wasn't being reported even though it was in the core build.

**Requester Info**
Jacob Hageman - NASA/GSFC

skliper:
**Describe the contribution**
Fix #795 - Adds all the currently possible source files for vxworks to coverage statistics and adds coverage tests

- Renamed sockaddr* structures to deconflict from structure name in os-impl-bsd-sockets.c to make stubbing out the OS structures easier.
- Compiling os-impl-bsd-sockets.c in coverage test with OS_NETWORK_SUPPORTS_IPV6 to cover that option also
- Renamed bsd-select-stubs.c to sys-select-stubs.c since it's really stubbing sys/select.h not the bsd implementation of select

**Testing performed**
Build and execute unit tests, passed

**Expected behavior changes**
None, but coverage now includes all currently possible files in vxworks build

**System(s) tested on**
 - Hardware: cFS Dev Server
 - OS: Ubuntu 18.04
 - Versions: cFS Bundle main + this commit

**Additional context**
Not full coverage yet on  os-impl-bsd-sockets.c (plan to keep improving prior to CCB)
Also will squash

**Third party code**
None

Jacob Hageman - NASA/GSFC


skliper:
@jphickey Is this along the lines of what you'd expect?

jphickey:
> @jphickey Is this along the lines of what you'd expect?

Yes!  This looks good!

The only comment is regarding the endian conversions - e.g. ntohl/s, etc - these are ones that I would consider falling back to "real" behavior - similar to memcpy/memset etc - if nothing is configured.  Having it just return 0 by default might cause problems for some code.  But in those cases I suppose someone can just be sure to set/register a return value in the test case.

skliper:
> ...ntohl/s...

I think I left those out since host could be big or little endian, and preference might be a hook that emulates the real target?  Just don't want testers to assume coverage testing w/ the test host endian is really appropriate in all cases.

jphickey:
I do agree that if there is any conditional which depends on output of an endian conversion, that the test case should explicitly set that output, rather than assuming the stub is going to work like the real one does.  So it's probably fine the way it is.

astrogeco:
[CCB:2021-02-24](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-February-24-2021) 
- Doesn't get to 100% coverage yet

skliper:
See #831 for follow-on to finish adding coverage

skliper:
@astrogeco squashed, ready for merge

PR #908, opened 2021-03-15
Fix #861, compile time assert for sockaddr size

### LINKED ISSUE #861: Use compile time assert on address length check in bind
Known at compile time:
https://github.com/nasa/osal/blob/ead5723f59ee49b36a36d6cb66c4cfe680ebbd68/src/os/portable/os-impl-bsd-sockets.c#L203

Use CompileTimeAssert

What if sizeof(struct sockaddr_in6) is > OS_SOCKADDR_MAX_LEN but only AF_INET is used?  Not really an error?  Maybe only really known if OS_NETWORK_SUPPORTS_IPV6 is not defined?

**Additional context**
None

**Requester Info**

jphickey:
**Describe the contribution**
OSAL provides an abstract buffer for socket addresses, independent of the underlying implementation.  The size of this buffer is
configurable by the user via compile-time options.

This adds a CompileTimeAssert to confirm that the size of this abstract buffer is large enough to store any of the enabled
address types. This also removes the need for runtime tests.

Fixes #861

**Testing performed**
Build and sanity test, run unit tests

**Expected behavior changes**
This change means that the `OS_SOCKADDR_MAX_LEN`  __must__ be configured large enough for all enabled address types.  For instance, ipv6 addresses are likely to be larger than ipv4 addresses in the implementation.  It is therefore possible to se OS_SOCKADDR_MAX_LEN large enough for ipv4 but not large enough for ipv6.

A configuration such as this used to work if the runtime code only used ipv4 addresses.  If runtime code used ipv6 address, it would fail at runtime.  With this change, a configuration such as this will fail to compile, and enforce that the abstract size is large enough for any/all enabled address types, regardless of what is actually used.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/src/os/portable/os-impl-bsd-sockets.c
+++ b/src/os/portable/os-impl-bsd-sockets.c
@@ -73,6 +73,15 @@ typedef union
 #endif
 } OS_SockAddr_Accessor_t;
 
+/*
+ * Confirm that the abstract socket address buffer size (OS_SOCKADDR_MAX_LEN) is
+ * large enough to store any of the enabled address types.  If this is true, the
+ * size of the above union will match OS_SOCKADDR_MAX_LEN.  However, if any
+ * implemention-provided struct types are larger than this, the union will be
+ * larger, and this indicates a configuration error.
+ */
+CompileTimeAssert(sizeof(OS_SockAddr_Accessor_t) == OS_SOCKADDR_MAX_LEN, SockAddrSize);
+
 /****************************************************************************************
                                     Sockets API
  ***************************************************************************************/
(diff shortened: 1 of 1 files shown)


skliper:
Disregard my comment at the CCB, doesn't make much sense.  I'm OK with ==, assumes no funny padding.

jphickey:
OK - I took a second look and realized that it indeed needs to be `==` ... if it was `>=` then it would defeat the purpose - this means that one or more of the `struct` members were bigger than the generic field.

The only possible issue is that if there is some extra compiler-added padding, you might have a situation where the size was technically "big enough" but still fails the assert, but then you just have to increase the size a bit to make it a multiple of alignment, which should be common practice anyway (no big deal).

PR #915, opened 2021-03-17
Fix #888, better return codes from OS_SymbolTableDump_Impl

### LINKED ISSUE #888: Report symEach errors in OS_SymbolTableDump_Impl
Errors not reported:
https://github.com/nasa/osal/blob/ead5723f59ee49b36a36d6cb66c4cfe680ebbd68/src/os/vxworks/src/os-impl-symtab.c#L262

Check for non-NULL, report error (write to file?), return error.

None

**Additional context**
None

**Requester Info**

jphickey:
**Describe the contribution**
Improve the error codes from this function.

- Introduces a new error `OS_ERR_OUTPUT_TOO_LARGE` if the size limit was insufficient (instead of `OS_SUCCESS`).
- Return `OS_ERROR` if an empty file was written - this likely indicates some fundamental issue with the VxWorks symbol table.
- Return `OS_ERR_NAME_TOO_LONG` if one of the symbol names was too long, (instead of generic `OS_ERROR`).

Improve unit test to check for/verify these responses.

Fixes #888

**Testing performed**
Build and sanity check, run all unit tests, confirm coverage of new cases.

**Expected behavior changes**
Better/more descriptive return codes if OS_SymbolTableDump_Impl does not do what is expected.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/src/os/inc/osapi-error.h
+++ b/src/os/inc/osapi-error.h
@@ -84,6 +84,7 @@ typedef char os_err_name_t[OS_ERROR_NAME_LENGTH];
 #define OS_ERR_STREAM_DISCONNECTED     (-37) /**< @brief Stream disconnected */
 #define OS_ERR_OPERATION_NOT_SUPPORTED (-38) /**< @brief Requested operation not support on supplied object(s) */
 #define OS_ERR_INVALID_SIZE            (-40) /**< @brief Invalid Size */
+#define OS_ERR_OUTPUT_TOO_LARGE        (-41) /**< @brief Size of output exceeds limit  */
 
 /*
 ** Defines for File System Calls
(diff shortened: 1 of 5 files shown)


jphickey:
Also note - I did go back and check OSAL 4.2.1a source code - this version also returned "OS_SUCCESS" if/when the size limit was reached/exceeded.  But I think the reason is because it was not trivial to return a value through the `symEach()` callback, so it simply didn't.  So I think it was implemented this way merely to be "bug-for-bug" compatible with the historical behavior.

Otherwise, I can't think of a good reason to return success if the file was actually truncated.....

astrogeco:
[CCB:2021-03-24](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-March-24-2021) APPROVED
- Return codes were vague
- Silently truncated files but returned a "success" code 
- Adds new error code "output too large"

PR #930, opened 2021-03-25
Fix #929, use test function for osal_id_t

### LINKED ISSUE #929: file api unit test uses osal_id_t as integer
**Describe the bug**
The type `osal_id_t` should not be used as an integer, and normal integer operations or comparisons should not be used with this type.  However it looks like a couple cases in file API test were recently added which are comparing this to a `uint32` type directly.

**To Reproduce**
Build OSAL with a type-safe osal_id_t, fails to build in `file_api_test.c`

**Expected behavior**
Build should succeed.

**Code snips**
Comparisons to uint32:
https://github.com/nasa/osal/blob/53f7f6167052a94de4d2de96c62dd71a63601654/src/tests/file-api-test/file-api-test.c#L179
https://github.com/nasa/osal/blob/53f7f6167052a94de4d2de96c62dd71a63601654/src/tests/file-api-test/file-api-test.c#L241

**System observed on:**
Ubuntu 20.04

**Additional context**
Should use `!OS_ObjectIdDefined()` inline function to test.

**Reporter Info**

jphickey:
**Describe the contribution**
Do not initialize or compare osal_id_t directly to integers, use the provided comparison function.

Fixes #929

**Testing performed**
Build and run unit tests

**Expected behavior changes**
System builds and runs again when using a type-safe/non-integer osal_id_t type.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/src/tests/file-api-test/file-api-test.c
+++ b/src/tests/file-api-test/file-api-test.c
@@ -175,7 +175,7 @@ void TestCreatRemove(void)
     /* try creating with file name too big, should fail */
     status = OS_OpenCreate(&fd, longfilename, OS_FILE_FLAG_CREATE | OS_FILE_FLAG_TRUNCATE, OS_READ_WRITE);
     UtAssert_True(status < OS_SUCCESS, "status after create file name too long = %d", (int)status);
-    UtAssert_True(!OS_ObjectIdDefined(fd), "fd == OS_OBJECT_ID_UNDEFINED");
+    UtAssert_True(!OS_ObjectIdDefined(fd), "fd(%lu) not defined", OS_ObjectIdToInteger(fd));
 
     /* try removing with file name too big. Should Fail */
     status = OS_remove(longfilename);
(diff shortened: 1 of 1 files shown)


skliper (approved):

src/tests/file-api-test/file-api-test.c
```diff
@@ -173,10 +173,9 @@ void TestCreatRemove(void)
     UtAssert_True(status == OS_SUCCESS, "status after remove max name length file = %d", (int)status);
 
     /* try creating with file name too big, should fail */
-    fd     = ~OS_OBJECT_ID_UNDEFINED;
     status = OS_OpenCreate(&fd, longfilename, OS_FILE_FLAG_CREATE | OS_FILE_FLAG_TRUNCATE, OS_READ_WRITE);
     UtAssert_True(status < OS_SUCCESS, "status after create file name too long = %d", (int)status);
-    UtAssert_UINT32_EQ(fd, OS_OBJECT_ID_UNDEFINED);
+    UtAssert_True(!OS_ObjectIdDefined(fd), "fd == OS_OBJECT_ID_UNDEFINED");
```
zanzaben, line 178: Can this print out the actual value of fd instead of just always printing that it equals Undefined.

zanzaben (changes requested):

src/tests/file-api-test/file-api-test.c
jphickey: Updated in 56cf1ec

zanzaben (approved):

PR #938, opened 2021-03-31
Fix #887, specify shell name in ShellOutputToFile

### LINKED ISSUE #887: OS_ShellOutputToFile_Impl in VxWorks is not thread safe
Filename is common:
https://github.com/nasa/osal/blob/ead5723f59ee49b36a36d6cb66c4cfe680ebbd68/src/os/vxworks/src/os-impl-shell.c#L96

At minimum document.  Consider refactor for shell safe implementation.  Although current cFS concept is to use a single app and even this should be avoided when possible (direct shell use is discouraged in production ops).

Deprecate/remove.

**Additional context**
None

**Requester Info**

jphickey:
**Describe the contribution**
Rather than using a default name, use a specific name so that it can be polled more safely.

Fixes #887

**Testing performed**
Build and run coverage tests
Also enable shell and run "shell-test" on MCP750

**Expected behavior changes**
No longer using a string pointer which may be invalid.

**System(s) tested on**
MCP750, vxworks 6.9

**Additional context**
Although this while loop seems to be working OK, the overall "shell-test" actually fails on the MCP750 because it looks like the shell does not have an "echo" function.  This seems to be a separate issue, not sure if/how this test ever worked.

### DIFF
--- a/src/os/vxworks/src/os-impl-shell.c
+++ b/src/os/vxworks/src/os-impl-shell.c
@@ -33,6 +33,7 @@
 #include "os-impl-io.h"
 #include "os-shared-shell.h"
 #include "os-shared-file.h"
+#include "os-shared-task.h"
 #include "os-shared-idmap.h"
 #include "os-shared-common.h"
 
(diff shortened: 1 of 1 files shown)


jphickey:
Submitted issue #939 about related shell issue discovered when attempting to test this.

jphickey:
Also worth noting - this assumes that VxWorks will honor and use the shell name specified.  

Technically the VxWorks documentation for `shellGenericInit()` says that the actual shell name _may_ be different than the passed in shell name.  But it appears that is only possible if using a single shared shell task, which is not the case on the MCP750.

astrogeco:
[CCB 2021-03-31](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-March-31-2021) APPROVED

PR #941, opened 2021-04-02
Fix #871, allow OSAL re-initialization

### LINKED ISSUE #691: Combine initialized and shutdown flags into one
The OSAL global state structure has two flags, one for "initialized" (set during OS_API_Init) and one for shutdown (set during OS_ApplicationShutdown).  Although similar in purpose, they are defined and work differently.

Should combine these into just one state variable, which should be 0 prior to initialization (i.e. such that the application loader provided by the OS will zero it before OSAL loads), then set to a nonzero value for normal runtime, and another special nonzero value for shutdown.

**Additional context**
Just a suggestion for cleaning up/simplifying what is already there...

Structure defined here:
https://github.com/nasa/osal/blob/9407cdf0be77521d850c587c0f749779c4656c3e/src/os/shared/inc/os-shared-common.h#L40-L67.

Using a full 32 bit value with all bits - even though there are only a few states - provides protection in case of a random bit flip.  At least in the case of an application shutdown, the entire system operation shouldn't be dependent on a single bit in memory.  This is why the current "shutdown flag" is not just a simple 0/1.  So this should be preserved, but it should be trivial to combine this with the "Initialized" field.

**Requester Info**

### LINKED ISSUE #871: Return success on multiple OS_API_Init calls
OS_API_Init writes error to the syslog and returns error if called multiple times.  Doesn't really cause any issues, could just write to the syslog and return success.
https://github.com/nasa/osal/blob/ead5723f59ee49b36a36d6cb66c4cfe680ebbd68/src/os/shared/src/osapi-common.c#L115-L119

Consider returning success

Leave as-is (future work)

**Additional context**
None

**Requester Info**

jphickey:
**Describe the contribution**
Replaces the separate "Initialized" and "Shutdown" flags with a single state flag.  This simplifies things and makes for a single source of truth for the state of OSAL globally.

Note this is not strictly necessary, but it will likely be required if the intent is to be able to run tests (and OS_API_Init()) multiple times without a reboot in the middle to reset the state.

Fixes #871 
Fixes #691 

**Testing performed**
Build and sanity check CFE
Run all unit tests

**Expected behavior changes**
- Multiple invocations of `OS_API_Init()` are allowed - subsequent calls can be ignored
- Deleting of any internal objects that did get created if `OS_API_Init()` fails (this leaves system in same state as when it started)
- Allows Re-initialization of OSAL after `OS_ApplicationShutdown()` - may be relevant when running unit tests several times without rebooting.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/src/os/posix/src/os-impl-console.c
+++ b/src/os/posix/src/os-impl-console.c
@@ -100,7 +100,7 @@ static void *OS_ConsoleTask_Entry(void *arg)
         local = OS_OBJECT_TABLE_GET(OS_impl_console_table, token);
 
         /* Loop forever (unless shutdown is set) */
-        while (OS_SharedGlobalVars.ShutdownFlag != OS_SHUTDOWN_MAGIC_NUMBER)
+        while (OS_SharedGlobalVars.GlobalState != OS_SHUTDOWN_MAGIC_NUMBER)
         {
             OS_ConsoleOutput_Impl(&token);
             sem_wait(&local->data_sem);
--- a/src/os/rtems/src/os-impl-console.c
+++ b/src/os/rtems/src/os-impl-console.c
@@ -123,7 +123,7 @@ static void OS_ConsoleTask_Entry(rtems_task_argument arg)
         local = OS_OBJECT_TABLE_GET(OS_impl_console_table, token);
 
         /* Loop forever (unless shutdown is set) */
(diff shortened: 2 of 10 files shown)


astrogeco:
[CCB:2021-04-07](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-April-07-2021) APPROVED
- VxWorks doesn't allow calling OS_API_Init without a reboot

PR #946, opened 2021-04-06
Fix #945, shell implementation on posix and rtems

### LINKED ISSUE #945: RTEMS fails to build with shell enabled
**Describe the bug**
The prototype for the OS_ShellOutputToFile_Impl() function has gotten stale and needs an update.  Unfortunately this isn't built as part of any regularly-run test routine.

**To Reproduce**
Change OSAL config to enable shell
Build on RTEMS

**Expected behavior**
Build should pass

**Code snips**
https://github.com/nasa/osal/blob/65c584feea04b4c77389667a1e30ae56b4da35ce/src/os/rtems/src/os-impl-shell.c#L55

**System observed on:**
RTEMS 4.11.3 target, Ubuntu 20.04 host.

**Reporter Info**

jphickey:
**Describe the contribution**
Due to other changes the shell implementation on RTEMS and POSIX failed to build under certain configurations.  

- On POSIX the check of `active_id` field was not valid.  Using the existing `OS_CloseAllFiles()` function does what is needed here and is much simpler.
- On RTEMS, the shell does not recognize the redirect operators.  This was basically cut and pasted from Linux/bash shell syntax and probably never worked on RTEMS, and there was no test to verify.  Now that "shell-test" exists, it confirms that this does not actually work.

For now, the broken impl file is simply deleted, and RTEMS will always use `os-impl-no-shell.c`.  In a future version, the implementation can be fixed properly.  

**Testing performed**
Build POSIX and RTEMS with shell enabled (OSAL_CONFIG_INCLUDE_SHELL=true)
Run "shell-test" and confirm behavior on POSIX (passes) and RTEMS (skips)

(Note that CFE does not directly use the OSAL shell function, but apps might; it is optional)

**Expected behavior changes**
Broken code is removed, so OSAL builds with `OSAL_CONFIG_INCLUDE_SHELL=true` on all platforms, but RTEMS will still result in OS_ERR_NOT_IMPLEMENTED always.

**System(s) tested on**
Ubuntu 20.04 (native)
RTEMS 4.11.3

**Additional context**
I briefly investigated what it would take to fix this on RTEMS properly.  The issue is that the `rtems_shell_init()` function attaches to a "device name" (typically /dev/console for interactive use), not individual input/output file descriptors.  So it appears to be a matter of creating a pseudoterminal-type device that "reads" from the script and writes to the file.  But it also appeared this was an area that changed a fair bit in RTEMS between 4.11 and 5.x release, so any solution would be unlikely to work on both, and this is why I decided its probably not worth pursuing for now.  (a stakeholder who wants the shell to work on RTEMS can always re-implement).

But since the existing code in `os-impl-shell.c` was very broken, it is just removed (not even useful as a basis for future; better to use posix or vxworks as basis instead).


skliper:
CCB:2021-04-14 - APPROVED

PR #948, opened 2021-04-07
Fix #944, add OS_API_Teardown

### LINKED ISSUE #197: Black Box Unit Tests Do Not Ensure Resources Are Cleaned Up Between Tests
Unit tests should utilize a SetUp and TearDown approach to ensure each test runs in an expected and known configuration.

The TearDown function needs to ensure that all test objects and resources have been deallocated and cleanup.

### LINKED ISSUE #944: OSAL should have an inverse function to OS_API_Init()
Once OSAL is initialized, normally it runs forever until rebooted, or the process exits in the case of Linux.  In that light, resources obtained during OS_API_Init() cannot directly/simply be released via the API - there is no opposite/inverse routine defined.

However under testing conditions, and for general completeness of API, it is useful to have an inverse function to tear down the API structures and return the system to the state it was prior to the `OS_API_Init()` call.

This can (_almost_) be done with the current OSAL API, but it must be done in two parts:

1. `OS_DeleteAllObjects()` will clean up any remaining user-instantiated objects in the tables.
2. `OS_ApplicationShutdown(true)` will cause any internal resources, such as the console utility task, to also self-exit.

The combination of these two basically leaves the system in a state similar to what it was before `OS_API_Init()` was run.  (its not perfect, but its close).

But its not ideal because a user shouldn't have to call two functions (one of which also has a parameter) to undo the init call.

Implement a `void OS_API_Teardown(void)` routine which would be just be a wrapper around these two currently-existing functions, to make it simpler to do this.

Importantly, being a void/void routine means that it can be easily used with unit tests and the `UtTest_AddTeardown()` routine.

Leave as-is

**Additional context**
Issue #197 which calls for something like this that unit tests can use.

**Requester Info**

jphickey:
**Describe the contribution**
This cleans up all OSAL resources as best as possible, ideally leaving the system in a state where OS_API_Init() may be invoked again.

Fixes #944
Fixes #197 

**Testing performed**
Build and run all unit tests, confirm correct behavior
Test in conjunction with fix for #197, confirm test can be run multiple times on VxWorks without reboot

**Expected behavior changes**
Adds OS_API_Teardown API to complement OS_API_Init

**System(s) tested on**
Ubuntu 20.04 (native)
MCP750 vxworks 6.9

### DIFF
--- a/src/os/inc/osapi-common.h
+++ b/src/os/inc/osapi-common.h
@@ -140,6 +140,33 @@ int32 OS_API_Init(void);
 
 /*-------------------------------------------------------------------------------------*/
 /**
+ * @brief Teardown/de-initialization of OSAL API
+ *
+ * This is the inverse of OS_API_Init().  It will release all OS resources and
+ * return the system to a state similar to what it was prior to invoking
+ * OS_API_Init() initially.
+ *
+ * Normally for embedded applications, the OSAL is initialized after boot and will remain
+ * initialized in memory until the processor is rebooted.  However for testing and
+ * developement purposes, it is potentially useful to reset back to initial conditions.
+ *
+ * For testing purposes, this API is designed/intended to be compatible with the
+ * UtTest_AddTeardown() routine provided by the UT-Assert subsystem.
+ *
+ * @note This is a "best-effort" routine and it may not always be possible/guaranteed
+ * to recover all resources, particularly in the case of off-nominal conditions, or if
+ * a resource is used outside of OSAL.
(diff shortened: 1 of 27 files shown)


jphickey:
Updated, also including the fix for issue #197 in this PR (second commit)

astrogeco:
[CCB:2021-04-07](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-April-07-2021) APPROVED

PR #949, opened 2021-04-07
Fix #197, test teardown cleanup

### LINKED ISSUE #197: Black Box Unit Tests Do Not Ensure Resources Are Cleaned Up Between Tests

jphickey:
**Describe the contribution**
Invokes OS_API_Teardown as a test teardown function for all unit tests.  This returns the global state objects to their initial state and therefore allows the test to be re-run/repeated without rebooting.  It also ensures that any resources allocated by the test are removed before the test finishes.

**Testing performed**
Build and run all unit tests, confirm passing

**Expected behavior changes**
Test resources are cleaned up to the extent possible, and set back to initial state.  Tests can be repeated if desired.

**System(s) tested on**
Ubuntu 20.04
MCP750 Vxworks 6.9

**Additional context**
Depends on #948 

### DIFF
--- a/src/os/inc/osapi-common.h
+++ b/src/os/inc/osapi-common.h
@@ -140,6 +140,33 @@ int32 OS_API_Init(void);
 
 /*-------------------------------------------------------------------------------------*/
 /**
+ * @brief Teardown/de-initialization of OSAL API
+ *
+ * This is the inverse of OS_API_Init().  It will release all OS resources and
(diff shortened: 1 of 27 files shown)


jphickey:
Pushed initially as draft since this depends on #948.

PR #971, opened 2021-04-27
Fix #969, socket accept using incorrect record

### LINKED ISSUE #969: Wrong token name used in accept call
https://github.com/nasa/osal/blob/afb5f7ba3ea48823e44490ce4258aa92a073087f/src/os/shared/src/osapi-sockets.c#L269

This is a bug. I think it should be conn_token.. not sock_token

jphickey:
**Describe the contribution**
This improves the network-api-test to do multiple connections, re-using the same acceptor socket between them.

As reported in #969, the OS_SocketAccept call was using the incorrect token, using the server/acceptor socket when it should have used the connection socket.

The bug overwrote data in the acceptor socket, but it would only cause an issue when the user attempted to use the server socket to accept a second connection, but the tests only performed a single connection.

With the improvement to network-api-test to run multiple connections, it now reveals the bug in the test.  Without the fix, the test failed on the second connection.  With the fix to OS_SocketAccept to use the correct token, the test passes fine again.

Fixes #969 

**Testing performed**
Build and sanity check CFE, run all unit tests (particularly network-api-test).

**Expected behavior changes**
OS_SocketAccept() does not clobber the server socket, allows it to be re-used for additional connection(s) as intended.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/src/os/shared/src/osapi-sockets.c
+++ b/src/os/shared/src/osapi-sockets.c
@@ -266,7 +266,7 @@ int32 OS_SocketAccept(osal_id_t sock_id, osal_id_t *connsock_id, OS_SockAddr_t *
             if (return_code == OS_SUCCESS)
             {
                 conn_record = OS_OBJECT_TABLE_GET(OS_global_stream_table, conn_token);
-                conn        = OS_OBJECT_TABLE_GET(OS_stream_table, sock_token);
+                conn        = OS_OBJECT_TABLE_GET(OS_stream_table, conn_token);
 
                 /* Incr the refcount to record the fact that an operation is pending on this */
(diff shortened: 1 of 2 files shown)


jphickey:
Re-Pushed to fix formatting.....

skliper (approved):

skliper:
Ping @excaliburtb

jphickey:
Updated this PR, I added a second commit which changes the explicit test of the count pattern to utilize the provided `UtMemFill` and `UtAssert_MemCmpCount` APIs, respectively.  No other tests were using this facility so this helps confirm that it works correctly.

astrogeco:
Do we want approval from @excalibur before merge?

PR #973, opened 2021-04-27
Fix #964, UtPrintx function

### LINKED ISSUE #964: Why does UtPrintx have an infinite loop in it?
**Describe the bug**
Any use of UtPrintx results in an infinite printing of the given Memory and Length.

**To Reproduce**
1. Put UtPrintx in a unit test, give it an address and size (from 0 to max uint32)
2. Run tests
3. Enter loop

**Expected behavior**
1. Print the number of bytes given as _Length_ starting at the given address _Memory_ (with newline at end)
2. Return

**Code snips**
https://github.com/nasa/osal/blob/b37da18bd5efed6d86344e4cafaa42c833b015fb/ut_assert/src/uttools.c#L188-L207

**System observed on:**
 - RHEL 7.8

**Reporter Info**
Alan Gibson
NASA/GSFC 587

jphickey:
**Describe the contribution**
Fix the UtPrintx() routine such that the loop stops correctly.  Also improves the output to print the address, not just the data.

Note if UtPrintf is used, one sees the file/line of the UtPrintx function, not the actual test location, so it is better to call
UT_BSP_DoText directly so it omits this extraneous info.

Fixes #964 

**Testing performed**
Temporarily Update a test to use UtPrintx() -- this function was not used in any current tests -- and confirm that the output now works as expected.

**Expected behavior changes**
No infinite loop in UtPrintx, and also it now includes the memory address of the data, not just the data itself.

**System(s) tested on**
Ubuntu 20.04

Joseph Hickey Vantage Systems, Inc.

### DIFF
--- a/ut_assert/src/uttools.c
+++ b/ut_assert/src/uttools.c
@@ -190,19 +190,22 @@ void UtPrintx(const void *Memory, uint32 Length)
     uint32       i;
     uint32       j;
     const uint8 *Byte_ptr = Memory;
-    char         OutputLine[50];
+    char         OutputLine[80];
     char *       OutPtr;
 
     i = 0;
-    while (1)
+    while (i < Length)
     {
+        snprintf(OutputLine, sizeof(OutputLine), "%16lx: ", (unsigned long)&Byte_ptr[i]);
         OutPtr = OutputLine;
+        OutPtr += strlen(OutputLine);
         for (j = 0; j < 16 && i < Length; j++, i++)
         {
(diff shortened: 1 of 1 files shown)


skliper (approved):

astrogeco:
[CCB:2021-04-28](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-April-28-2021) APPROVED

jphickey:
The branch was deleted by mistake, this is still needing merge

PR #980, opened 2021-05-04
Fix #331, scrub return values

### LINKED ISSUE #331: Assert return codes specified in API in functional tests
I noticed that osal-core-test.c only tests if the function returns OS_SUCCESS or does NOT return OS_SUCCESS.  It does not actually test that the return code is correct.  For example, the test to ensure that the OSAL does not create a queue with a name that already exists does not actually test for OS_ERR_NAME_TAKEN.  It merely passes if the return is not OS_SUCCESS.  A quick spot check indicates this design pattern is systemic.

These unit tests were used to certify (per NPR7150A) the ARINC 653 version of CFS.  This requires requirements have tracing back to tests.  Testing the actual expected return code is a better pattern, but is required if there is an actual requirement defining the return codes.

jphickey:
**Describe the contribution**
Confirm consistency between the return values generated by each OSAL API and the doxygen documentation.  Each explicitly-returned status code should have a corresponding "retval" entry in the documentation.

Also confirms that each explicitly-returned status code has a matching case in the coverage tests that specifically checks for
that return value.  Some were actually missing, where as some were just being reported incorrectly.

Finally this also corrects one argument name mismatch in QueueCreate where it was prototyped as "data_size" but implemented as "max_size".

Fixes #331

**Testing performed**
Build and sanity check CFE, run all unit tests, build osalguide documentation and confirmed no warnings

**Expected behavior changes**
None, test and documentation updates only.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/src/os/inc/osapi-clock.h
+++ b/src/os/inc/osapi-clock.h
@@ -83,6 +83,7 @@ enum
  * @param[out]  time_struct An OS_time_t that will be set to the current time
  *
  * @return Get local time status, see @ref OSReturnCodes
+ * @retval #OS_INVALID_POINTER if time_struct is null
  */
 int32 OS_GetLocalTime(OS_time_t *time_struct);
 
@@ -97,6 +98,7 @@ int32 OS_GetLocalTime(OS_time_t *time_struct);
  * @param[in]  time_struct An OS_time_t containing the current time
  *
  * @return Set local time status, see @ref OSReturnCodes
+ * @retval #OS_INVALID_POINTER if time_struct is null
  */
 int32 OS_SetLocalTime(const OS_time_t *time_struct);
 
--- a/src/os/inc/osapi-common.h
+++ b/src/os/inc/osapi-common.h
@@ -162,8 +162,6 @@ int32 OS_API_Init(void);
  * For example, while this will attempt to unload all dynamically-loaded modules, doing
  * so may not be possible and/or may induce undefined behavior if resources are in use by
  * tasks/functions outside of OSAL.
- *
- * @return None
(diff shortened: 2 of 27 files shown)


jphickey:
Note this currently just focuses on the coverage test aspect - i.e. that every return value _directly_ generated by an API is both documented and tested for in the coverage test.

astrogeco:
[CCB:2021-05-05](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-May-05-2021) 
- Scrubs return codes
- Replace "Fix" with "Partial" or "Related"

astrogeco:
> Needs a follow-on to add all of our standard input checking (even for pass through return codes) along with the black box tests. Probably would benefit from documenting the list - NULL pointer checks (if enabled), filename/path validity checks, etc.

Should we open a new issue for these?

Also, did we open the issue in cFE to do the same scrub? 
> CCB:2021-05-05, Open cFE ticket to do the same thing

skliper:
> > Needs a follow-on to add all of our standard input checking (even for pass through return codes) along with the black box tests. Probably would benefit from documenting the list - NULL pointer checks (if enabled), filename/path validity checks, etc.
> 
> Should we open a new issue for these?
> 
> Also, did we open the issue in cFE to do the same scrub?
> 
> > CCB:2021-05-05, Open cFE ticket to do the same thing

#331 is still open to complete the follow-on in OSAL.  Note the tests are still being added in cFE, and have associated open issues.  We just need to confirm they check the necessary return codes (follow the same patter as OSAL).

PR #987, opened 2021-05-06
Fix #986, Show CodeQL Preview

### LINKED ISSUE #986: Show CodeQL Preview
Code preview is not available for code scanning results from CodeQL. 

The preview is available when the osal submodule is checked out rather than the cFS bundle. I have not found a way to checkout just the submodule AND use the tests. 

Currently, the workflow checks out the cFS bundle. 
```
    steps:
      - name: Checkout bundle
        uses: actions/checkout@v2
        with:
          repository: nasa/cFS
          submodules: true

      - name: Checkout submodule
        uses: actions/checkout@v2
        with:
          path: osal
```
To show the preview, only the osal submodule needs to be checked out:
```
      - name: Checkout submodule
        uses: actions/checkout@v2
```

Would like to keep the tests, but believe they require the cFS bundle. 

**Additional context**
Community Support Thread: https://github.community/t/codeql-with-submodules/159839/10
> The reason is that we currently don’t look for files across submodules for the preview.

**Requester Info**
Ariel Adams, ASRC Federal

ArielSAdamsNASA:
**Describe the contribution**
Fix #986 
Checkout just the osal repository rather than the entire cFS bundle to allow CodeQL to provide a code preview. 
Code preview is not available when the entire cFS bundle is checked out. The osal itself must be the only repository checked out. 
Since the tests require the cFE repo, they are removed.  

**Testing performed**
Tested locally on forked repository 
image

Tested on this PR. Example of code scanning result found here: https://github.com/nasa/osal/security/code-scanning/142?query=ref%3Arefs%2Fpull%2F987%2Fmerge
**Expected behavior changes**
Code preview should be available in the osal repository. 

**Additional context**
Tests are not used. 

Ariel Adams, ASRC Federal

### DIFF
--- a/.github/workflows/codeql-osal-default.yml
+++ b/.github/workflows/codeql-osal-default.yml
@@ -1,0 +1,57 @@
+name: "CodeQL Analysis OSAL Default Build"
+
+on:
+  push:
+  pull_request:
+
+env:
+  SIMULATION: native
+  ENABLE_UNIT_TESTS: true
+  OMIT_DEPRECATED: true
+  BUILDTYPE: release
+  PERMISSIVE_MODE: true
+
+jobs:
+
+  #Checks for duplicate actions. Skips push actions if there is a matching or duplicate pull-request action. 
+  check-for-duplicates:
+    runs-on: ubuntu-latest
+    # Map a step output to a job output
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@master
+        with:
+          concurrent_skipping: 'same_content'
+          skip_after_successful_duplicate: 'true'
+          do_not_skip: '["pull_request", "workflow_dispatch", "schedule"]'
+
+
+  CodeQL-Build:
+    #Continue if check-for-duplicates found no duplicates. Always runs for pull-requests.
+    needs: check-for-duplicates
+    if: ${{ needs.check-for-duplicates.outputs.should_skip != 'true' }}
+    runs-on: ubuntu-18.04
+    timeout-minutes: 15
+
+    steps:
+      - name: Checkout submodule
+        uses: actions/checkout@v2
+
+      - name: Initialize CodeQL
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-build.yml
```diff
@@ -39,15 +28,11 @@ jobs:
 
       - name: Set up for build
         run: |
-          cp ./cfe/cmake/Makefile.sample Makefile
-          cp -r ./cfe/cmake/sample_defs sample_defs
-          make prep
-
+          cp Makefile.sample Makefile
+          make ENABLE_UNIT_TESTS=true PERMISSIVE_MODE=true prep
```
skliper, line 32: Recommend either using all env settings or all specified on the make line, looks like at minimum ENABLE_UNIT_TESTS is set in both locations.

skliper (commented):
One impact of this change is now the analysis is run using the osal default configuration, vs what is specified in cFE.  Probably OK as long as the cFE still does CodeQL, but then need to keep aware of OSAL warnings in that context.  Eventually we probably want at least 2 CodeQL analysis runs in OSAL to exercise a range of options.

.github/workflows/codeql-build.yml
ArielSAdamsNASA: Set PERMISSIVE_MODE in env settings and remove ENABLE_UNIT_TESTS in make command.

ArielSAdamsNASA:
> One impact of this change is now the analysis is run using the osal default configuration, vs what is specified in cFE. Probably OK as long as the cFE still does CodeQL, but then need to keep aware of OSAL warnings in that context. Eventually we probably want at least 2 CodeQL analysis runs in OSAL to exercise a range of options.

Renamed newly created CodeQL workflow as codeql-default-build to allow two analysis runs of CodeQL.

ArielSAdamsNASA:
Renamed workflow to codeql-osal-default.

ArielSAdamsNASA:
> Should we add the "check for duplicates" workflow to this?

Yes! Thanks for the reminder, just added it.

PR #990, opened 2021-05-07
Fix #982, Add test for object id inline functions

### LINKED ISSUE #982: Add tests for object id inline functions
`OS_ObjectIdToInteger`, `OS_ObjectIdFromInteger`, `OS_ObjectIdEqual`, `OS_ObjectIdDefined` are missing tests.

Add tests.  

Since these are inline they can be completely exercised via explicit coverage tests (suggest adding to coveragetest-idmap.c).  

Note these can be trivial, just need to convert a to and from a range of ID's (note these are just to/from integers, NOT indexes...), check each for equal to itself but not equal to the others, check that OS_OBJECT_ID_UNDEFINED results in undefined and the rest are defined.  This also ensures an ID didn't "wrap" by exercising Array Index conversions from array index 0 and max for each type (as defined in configuration).

Could be added to a "functional/full stack" test but no point since they don't have any external dependencies.

**Additional context**
None

**Requester Info**
Jacob Hageman - NASA/GSFC

zanzaben:
**Describe the contribution**
Fixes #982 
Adds test for inline object id functions

**Testing performed**
Build and run unit test

**Expected behavior changes**
No impact to behavior

**System(s) tested on**
Ubuntu 20.04

Alex Campbell GSFC


src/unit-test-coverage/shared/src/coveragetest-idmap.c
```diff
@@ -1065,6 +1070,76 @@ void Test_OS_ObjectIdIterator(void)
...
+    uint32            maxType      = 0;
+    uint32            typeReturn   = 0;
+    osal_objtype_t    idtype;
+    char              str[maxRecords];
+
+    for (idtype = 0; idtype < OS_OBJECT_TYPE_USER; ++idtype)
+    {
+        typeReturn = OS_GetMaxForObjectType(idtype);
+        maxType += typeReturn;
+    }
```
jphickey, line 1094: What is the philosophy here with adding up all the max values?  Why not check each one individually?  For instance if there was a bug that caused two values to get swapped, this would not catch it, because it would still add up to the same result.

src/unit-test-coverage/shared/src/coveragetest-idmap.c
```diff
@@ -1065,6 +1070,76 @@ void Test_OS_ObjectIdIterator(void)
...
+
+    if (maxType == maxRecords)
+    {
+        for (idtype = 0; idtype < OS_OBJECT_TYPE_USER; ++idtype)
+        {
+            actual = OS_SUCCESS;
+            do
+            {
+                sprintf(str, "%d", recordscount);
```
jphickey, line 1103: Should always use snprintf(), even if the buffer is known to be long - because in a future change someone can chance the definition of "str".

src/unit-test-coverage/shared/src/coveragetest-idmap.c
```diff
@@ -1065,6 +1070,76 @@ void Test_OS_ObjectIdIterator(void)
...
+
+                    recordscount++;
+                }
+
+            } while (actual == OS_SUCCESS);
+        }
+
+        for (int i = 0; i < recordscount; i++)
+        {
+            UtAssert_True(OS_ObjectIdDefined(typesI[i]), "%d Is defined", typesI[i]);
```
jphickey, line 1120: Needs %lu format code and use "OS_ObjectIdToInteger()" to print.  Should not attempt to print osal_id_t directly - it might not be integer.

src/unit-test-coverage/shared/src/coveragetest-idmap.c
```diff
@@ -1065,6 +1070,76 @@ void Test_OS_ObjectIdIterator(void)
...
+
+            for (int j = 0; j < recordscount; j++)
+            {
+                if (i == j)
+                {
+                    UtAssert_True(OS_ObjectIdEqual(typesI[i], typesJ[j]), "%d equals %d", typesI[i], typesJ[j]);
+                }
+                else
+                {
+                    UtAssert_True(!OS_ObjectIdEqual(typesI[i], typesJ[j]), "%d does not equal %d", typesI[i],
```
jphickey, line 1130: Both of these also need OS_ObjectIdToInteger

jphickey (changes requested):
Some comments noted inline.

I would also suggest just doing one record of each type.  "maxing it out" by calling OS_ObjectIdAllocateNew until it fails seems like overdoing it.  These inline functions are just simple ops and generally have no conditionals, there isn't a real benefit to testing it with every possible value.

src/unit-test-coverage/shared/src/coveragetest-idmap.c
skliper: Part of it is to confirm the total calculation in the test matches the calculation in code, since there isn't a total available externally, and these should never get out of sync.  Alternative would be to expose the total.

skliper:
I prefer maxing out for the comparison test.  Would catch any strange rollover bugs, or bug in creating the unique IDs.  This isn't a huge number, so seems reasonable for the additional "black box" coverage to ensure uniqueness.

jphickey:
> I prefer maxing out for the comparison test. Would catch any strange rollover bugs, or bug in creating the unique IDs. This isn't a huge number, so seems reasonable for the additional "black box" coverage to ensure uniqueness.

That's done by the existing test cases for `OS_ObjectIdAllocateNew` though - this calls it `OS_OBJECT_INDEX_MASK` times (plus some extra) to ensure the rollover happens correctly.

If the objective of these tests should be to check the inline functions, just a few choice values should be enough (I'd consider adding `OS_OBJECT_ID_UNDEFINED` too, which is not part of the current set).  They are really just type-safety tools at the end of the day (typecast replacement and a type-specific equality check) and don't have any real logic in them.

src/unit-test-coverage/shared/src/coveragetest-idmap.c
zanzaben: Fixed

src/unit-test-coverage/shared/src/coveragetest-idmap.c
zanzaben: fixed

src/unit-test-coverage/shared/src/coveragetest-idmap.c
zanzaben: fixed

skliper:
I still don't see any harm to confirm ID's across all the types are never equal except to themselves.  It adds very little complexity to the test.  The test for equality after the conversion from/to index across all IDs that can be allocated also shows the conversion doesn't change anything or cause loss of info.  Why skip when there's such little "cost"?  Plus it's already implemented... so is it really worth backing out a more thorough test?  I don't really care that it may be caught by some other test somewhere based on how it's implemented, that other test doesn't also exercise the conversions so I'd rather keep the full range i here.  Also avoids dependencies on other tests (say the other test changes, could make a reduced test here now miss something else).

astrogeco:
[CCB:2021-05-12](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-May-12-2021) CHANGES REQUESTED
- Don't like repeating things in tests 
- Open issue to move API to header? This can help to expose for the coverage test. 
- Add a test against "undefined".
- The problem is that we don't test these in functional tests 
- If time is trivial, it's okay to overtest a little
- Watch out for log growing too big in the future or tests taking too long
-

astrogeco:
@zanzaben do we want to change this to draft since there more work to do?

zanzaben:
All requested changes have been made.

src/unit-test-coverage/shared/src/coveragetest-idmap.c
```diff
@@ -1065,6 +1065,76 @@ void Test_OS_ObjectIdIterator(void)
...
+            if (actual == OS_SUCCESS)
+            {
+                typesI[recordscount] = token.obj_id;
+                intID                = OS_ObjectIdToInteger(typesI[recordscount]);
+                typesJ[recordscount] = OS_ObjectIdFromInteger(intID);
+
+                recordscount++;
+            }
+
+        } while (actual == OS_SUCCESS);
```
jphickey, line 1101: The other thing I'd like to see as part of this loop is a check that `recordscount` does not exceed `OS_MAX_TOTAL_RECORDS`.  It shouldn't when things are working right.... but should confirm that.

src/unit-test-coverage/shared/src/coveragetest-idmap.c
```diff
@@ -1065,6 +1065,76 @@ void Test_OS_ObjectIdIterator(void)
...
+    int32             recordscount = 0;
+    osal_objtype_t    idtype;
+    char              str[OS_MAX_TOTAL_RECORDS];
+
+    for (idtype = 0; idtype < OS_OBJECT_TYPE_USER; ++idtype)
+    {
+        actual = OS_SUCCESS;
+        do
+        {
+            snprintf(str, sizeof(str), "%d", recordscount);
```
jphickey, line 1088: This needs to be `(int)recordscount`

jphickey (changes requested):
Getting better, but still a few more concerns.

src/unit-test-coverage/shared/src/coveragetest-idmap.c
zanzaben: fixed

src/unit-test-coverage/shared/src/coveragetest-idmap.c
zanzaben: fixed

src/unit-test-coverage/shared/src/coveragetest-idmap.c
```diff
@@ -1065,6 +1065,76 @@ void Test_OS_ObjectIdIterator(void)
...
+     * OS_ObjectIdToInteger, OS_ObjectIdFromInteger, OS_ObjectIdEqual, OS_ObjectIdDefined
+     */
+    int32             actual;
+    OS_object_token_t token;
+    osal_id_t         typesI[OS_MAX_TOTAL_RECORDS];
+    osal_id_t         typesJ[OS_MAX_TOTAL_RECORDS];
+    uint32            intID;
+    int32             recordscount = 0;
+    osal_objtype_t    idtype;
+    char              str[OS_MAX_TOTAL_RECORDS];
```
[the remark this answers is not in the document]
zanzaben: fixed

jphickey (approved):

jphickey:
I merged this into my branch and ran it, looks OK ....

**BUT** - I still think we should reconsider testing _all_ the values.  In my test this produced 79243 test cases, which all get written to the log file.

Yes, it runs pretty fast on Linux where the console is local.  But the log file ends up being almost 6MB in size.  If running this on VxWorks on a 9600 baud serial connection, this will likely take _1-2 hours_ to print all those messages to the console.

IMO no test should produce more than 1000 cases (there are 3 digits allocated in the template).  Even that is a little excessive.

Can we just limit it to 5 IDs of each type?

src/unit-test-coverage/shared/src/coveragetest-idmap.c
```diff
@@ -1065,6 +1065,76 @@ void Test_OS_ObjectIdIterator(void)
...
+        for (int j = 0; j < recordscount; j++)
+        {
+            if (i == j)
+            {
+                UtAssert_True(OS_ObjectIdEqual(typesI[i], typesJ[j]), "%lu equals %lu", OS_ObjectIdToInteger(typesI[i]),
+                              OS_ObjectIdToInteger(typesJ[j]));
+            }
+            else
+            {
+                UtAssert_True(!OS_ObjectIdEqual(typesI[i], typesJ[j]), "%lu does not equal %lu",
```
jphickey, line 1117: This line is what produces the majority of the 79243 cases and makes the log huge.

jphickey:
Alternatively - if we can switch the "does not equal" test (noted in my previous comment) to only assert if it fails.  In other words, something like

```
if (ObjectIdEqual(...)) 
{
    UtAssert_Failed(....)
}
```

astrogeco:
@zanzaben and @jphickey are we good to go with this one?

PR #992, opened 2021-05-07
Fix #989, add local mutex to BSP console

jphickey:
**Describe the contribution**
Adds a generic mutex facility to the low level BSP.  This is mainly for use with low level BSP console output, but could be used for other items that should only be accessed by one task at a time.

This needs to be actually implemented in the BSP layer, so it can be used by UtAssert report outputs (pass/fail message) in addition to OS_printf().  Note that UtAssert always uses the OSAL BSP and does its text output via this layer, but it does __not__ necessarily use the full OSAL with its OS_printf/console abstraction.  So by implementing the lock in the BSP, it works with all modes of operation.

This also converts `OS_DEBUG` outputs to use the same console output facility rather than calling `fprintf()` directly.

The combination of a mutex and all common print outputs (UtAssert, OS_printf, OS_DEBUG) going through the same
path makes the test output on VxWorks much more coherent.

**Testing performed**
Build and sanity check CFE and run all unit tests on native and RTEMS.  Also build and run select tests on VxWorks to confirm.

**Expected behavior changes**
Console Output when running UtAssert-based tests is not garbled, much more coherent and readable.

**System(s) tested on**
Ubuntu 20.04 (native)
RTEMS 4.11 (QEMU)
VxWorks 6.9 (MCP750)

**Contributor Info - All information REQUIRED for consideration of pull request**
Joseph Hickey, Vantage Systems, Inc.

PR #993, opened 2021-05-07
Fix #988, do not require nonblock mode

### LINKED ISSUE #988: "network-api-test" does not run successfully on VxWorks
**Describe the bug**
Running the "network-api-test" on VxWorks fails with several issues.  The fcntl `F_GETFL`/`F_SETFL` calls return errno `EINVAL` followed by other failures.

**To Reproduce**
Run the "network-api-test" program on MCP750 w/VxWorks 6.9.  Various failures are reported.  Eventually the test hangs because a timeout test is reached, but timeouts do not function correctly due to issues with the previous fcntl F_GETFL calls.

**Expected behavior**
Test should pass.

**System observed on:**
MCP750 w/VxWorks 6.9

**Additional context**
This is partly related to modifications in PR #910/issue #863 which was a product of the CFS-38 review.  Previously it would not check the result of the `fcntl` calls so the "selectable" flag was always set.  Now the result is checked and the "selectable" flag only gets set if the flags were set.  So the fcntl was always failing on VxWorks but it was previously ignored.  But even if the O_NONBLOCK flag doesn't get set, `select()` should still be usable.  The flag is mainly set as a protection in case two tasks call a blocking API at the same time.

**Reporter Info**
Joseph Hickey, Vantage Systems, Inc.  (on behalf of JSC team)

jphickey:
**Describe the contribution**
In some versions of VxWorks the fcntl F_GETFL/F_SETFL opcodes do not appear to be implemented, and thus it is not possible
to set O_NONBLOCK mode.  However, this mode is not necessarily required, it is more of a backup/failsafe.

The "selectable" flag should not be dependent on whether O_NONBLOCK flag got set.

This also adjust some timeouts and adds some delays to improve the reliability of network-api-test on VxWorks.  The timeouts
were only 10ms, and this is much too short as messages are getting written on a 9600 baud console (avg 1ms/char).  A single log message can easily take 50-60ms alone.

Fixes #988 

**Testing performed**
Execute network-api-test on supported platforms (esp. VxWorks 6.9 in particular)

**Expected behavior changes**
Test now passes reliably.

**System(s) tested on**
Ubuntu 20.04 (native)
VxWorks 6.9 (MCP750)

**Additional context**
This also requires/depends on #992 for reliable operation.  The console mutex is required to ensure that the various information messages from the client/server tasks inside network-api-test do not clobber eachother, but it also helps synchronize them too.

### DIFF
--- a/src/os/portable/os-impl-bsd-sockets.c
+++ b/src/os/portable/os-impl-bsd-sockets.c
@@ -155,6 +155,10 @@ int32 OS_SocketOpen_Impl(const OS_object_token_t *token)
      * Set the standard options on the filehandle by default --
      * this may set it to non-blocking mode if the implementation supports it.
      * any blocking would be done explicitly via the select() wrappers
+     *
+     * NOTE: The implementation still generally works without this flag set, but
+     * nonblock mode does improve robustness in the event that multiple tasks
+     * attempt to accept new connections from the same server socket at the same time.
      */
     os_flags = fcntl(impl->fd, F_GETFL);
     if (os_flags == -1)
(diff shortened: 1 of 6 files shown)


astrogeco:

PR #1000, opened 2021-05-11
Fix #999, propagate status code in OS_rmdir

### LINKED ISSUE #999: Return value of "OS_rmdir" implementation not being passed through
**Describe the bug**
The return value of `OS_DirRemove_Impl()` (the low level implementation) is not getting propagated back to the caller of `OS_rmdir()` as it should be.

**To Reproduce**
Call `OS_rmdir()` on a directory that shouldn't be removable, e.g. if it is not empty.  The `OS_DirRemove_Impl()` function fails and returns `OS_ERROR`, but the application gets returned `OS_SUCCESS`.

**Expected behavior**
Caller should get the error code, not `OS_SUCCESS`.

**Code snips**
https://github.com/nasa/osal/blob/706f0de6902eb442563b44ae58904100a0e72cd3/src/os/shared/src/osapi-dir.c#L251

**System observed on:**
Ubuntu 20.04

**Additional context**
Noticed this when validating return codes in #331 and #981

**Reporter Info**

jphickey:
**Describe the contribution**
The status code from the low level implementation needs to be returned from the caller, in case it was not OS_SUCCESS.

Fixes #999 

**Testing performed**
Build and sanity check CFE, run all tests

**Expected behavior changes**
The status returned from `OS_rmdir()` should now be correct if the implementation failed.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/src/os/shared/src/osapi-dir.c
+++ b/src/os/shared/src/osapi-dir.c
@@ -248,7 +248,7 @@ int32 OS_rmdir(const char *path)
     return_code = OS_TranslatePath(path, local_path);
     if (return_code == OS_SUCCESS)
     {
-        OS_DirRemove_Impl(local_path);
+        return_code = OS_DirRemove_Impl(local_path);
     }
 
     return return_code;


skliper (approved):

astrogeco:

PR #1001, opened 2021-05-12
Fix #981, rework "unit-tests" to use macros

### LINKED ISSUE #981: Improve consistency in the result reporting in "unit-tests"
In order to be able to cross check and confirm that the unit tests are actually checking for the status codes that they should be, the test cases need to be clear about the API being called and the return value/status code being tested for in that test case.  However, the functional tests are currently not consistent in how results are reported.

Example from the binsem test, which is better because it at least uses a macro that includes both the function name/API being called (`OS_BinSemCreate`) as well as the status code being tested for (`OS_INVALID_POINTER`):

https://github.com/nasa/osal/blob/a061666924644a7d23b17d3d060dc9e704913447/src/unit-tests/oscore-test/ut_oscore_binsem_test.c#L84

Which translates to a line like this in the log file:
`[ PASS] 05.002 ut_oscore_binsem_test.c:84 - OS_BinSemCreate(&sem_ids[0], NULL, 1, 0) (-2) == OS_INVALID_POINTER (-2): null pointer arg 2`

However a similar test case from OS_ModuleLoad displays neither:
https://github.com/nasa/osal/blob/a061666924644a7d23b17d3d060dc9e704913447/src/unit-tests/osloader-test/ut_osloader_module_test.c#L99-L105

The latter translates only to a line like this in the log file, which doesn't say much at all:
`[ PASS] 01.002 ut_osloader_module_test.c:103 - #2 Null-pointer-arg-2`

**Additional context**
This is a prerequisite to completing #331 - The information in the log file needs to be made more like the BinSem case, so it can be cross referenced back to confirm the functional tests are checking for the items they are supposed to be checking for.

**Requester Info**

jphickey:
**Describe the contribution**
Update most of the conditions in the "unit tests" to use the support macros whenever possible.  The macros display the
function being called as well as the return value being checked for, and therefore this makes it so the test log file can be
cross referenced back to the documentation to ensure the documented return codes are being tested.

As a nice side effect this also removes quite a bit of repetition in the test programs.

Fixes #981

**Testing performed**
Build and run all OSAL unit tests

**Expected behavior changes**
Log file produced has much more detail, and can now be cross referenced back to documentation.

**System(s) tested on**
Ubuntu 20.04 (native)
RTEMS 4.11 (pc686+qemu)

**Additional context**
See gist of complete test output log for "osal_core_UT": https://gist.github.com/jphickey/ccb739d1f1cf6b36caca73e5ee9205cf
Dependency note: PR #1000 fixes the OS_rmdir test.

### DIFF
--- a/src/unit-tests/inc/ut_os_support.h
+++ b/src/unit-tests/inc/ut_os_support.h
@@ -61,42 +61,59 @@
  */
 #define UT_OS_IO_BUFF_SIZE 128
 
-static inline bool UtOsalRetVal(int32 Fn, int32 Exp, UtAssert_CaseType_t casetype, const char *File, uint32 Line,
-                                const char *FnTxt, const char *ExpTxt, const char *Msg)
+static inline bool UtOsalRetVal(int32 Fn, int32 Exp, bool NotImplAllowed, UtAssert_CaseType_t casetype,
+                                const char *File, uint32 Line, const char *FnTxt, const char *ExpTxt)
 {
-    return UtAssertEx(Fn == Exp, casetype, File, Line, "%s (%d) == %s (%d): %s", FnTxt, (int)Fn, ExpTxt, (int)Exp, Msg);
+    /* If "not implemented" is acceptable, override the casetype to be N/A */
+    if (NotImplAllowed && (Fn == OS_ERR_NOT_IMPLEMENTED || Fn == OS_ERR_OPERATION_NOT_SUPPORTED))
+    {
+        casetype = UTASSERT_CASETYPE_NA;
+    }
+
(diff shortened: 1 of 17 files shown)


astrogeco:
- Basically a rewrite of most of the unit tests
- Most of the tests were not using the new macros
- Adds new macros like `UT_Nominal_OR_NotImpl`

astrogeco:
@jphickey see test failures,

```

jphickey:
> @jphickey see test failures,
> 
> ```
> The following tests FAILED:
> 	 72 - osal_file_UT (Failed)
> Makefile:125: recipe for target 'test' failed
> make: *** [test] Error 8
> Error: Process completed with exit code 2.
> ```

Yes, as noted in the summary, this should be fixed by PR #1000 ( it is actually a FSW bug that wasn't detected by tests until now ).

PR #1028, opened 2021-05-19
Fix #1027, defer cancellation when BSP locked

### LINKED ISSUE #1027: Occasional deadlock issue in tests that delete tasks
**Describe the bug**
When running the unit tests repeatedly, occasionally some tests are getting into a deadlock.  These tests are ones that:

1. use sub-tasks to do various work
2. those sub-tasks use OS_printf()
3. use an asynchronous OS_TaskDelete during their cleanup/teardown

In the event that the sub-task was in the midst of an OS_printf() call when OS_TaskDelete was invoked, the underlying BSP lock will not get released.

Observed in mutex-test, but others may have similar patterns.

**To Reproduce**
Run mutex-test repeatedly, may deadlock at some runs.  (it is a race condition, not 100% reproducible)

**Expected behavior**
Should run consistently.

**System observed on:**
Ubuntu

**Additional context**
This really just a symptom of a generic/known issue with OS_TaskDelete, in that other resources held by that task are not necessarily tracked or freed, depending on what it was doing at the time it was deleted.

Linux/Pthreads does have a workaround but the issue is likely to exist on all OS's

**Reporter Info**

jphickey:
**Describe the contribution**

Resolves two related issues:
- OS_TaskGetId does not return a valid value for tasks where cancellation is pending, but they are still running.  This in turn is likely to trigger other (bogus) debug checks which invoke OS_DEBUG and in turn do console writes.
- The console write itself is a cancellation point, which is now done while holding a BSP mutex.  If canceled here, then the mutex is not released.

Solution is in two parts:
- OS_TaskGetId should return the task ID it knows about, regardless of whether the task is pending cancellation or not.
- Defer cancellation of the task while the BSP is locked, ensure it reaches the unlock, then restore the previous cancel state.

Fixes #1027 

**Testing performed**
Run mutex-test repeatedly, ensure no deadlocks

**Expected behavior changes**
No chance of deadlock, and also fixes a bogus debug message about unlocking from the wrong task if the task is pending delete.

**System(s) tested on**
Ubuntu


skliper (approved):

skliper:
@astrogeco requesting fastrack, bugs at this point have a high impact.

PR #1037, opened 2021-05-20
Fix #1021, resolve discrepancies between task API and unit tests

### LINKED ISSUE #1021: Unit test status code discrepancies in osapi-task.h
The unit tests and API documentation should match, where all status codes in the API documentation are tested in a unit test, and likewise all codes being tested for are documented in the API.  

Resolve the following mismatches:

File `src/os/inc/osapi-task.h`, function `OS_TaskCreate()`:
- [x] ONLY IN DOXY: OS_ERROR
- [x] ONLY IN DOXY: OS_ERR_INVALID_PRIORITY
- [x] ONLY IN DOXY: OS_ERR_INVALID_SIZE
- [x] OK: OS_ERR_NAME_TAKEN
- [x] OK: OS_ERR_NAME_TOO_LONG
- [x] OK: OS_ERR_NO_FREE_IDS
- [x] OK: OS_INVALID_POINTER
- [x] OK: OS_SUCCESS

File `src/os/inc/osapi-task.h`, function `OS_TaskDelete()`:
- [x] ONLY IN DOXY: OS_ERROR
- [x] OK: OS_ERR_INVALID_ID
- [x] OK: OS_SUCCESS

File `src/os/inc/osapi-task.h`, function `OS_TaskInstallDeleteHandler()`:
- [x] ONLY IN TEST: OS_ERR_INVALID_ID

File `src/os/inc/osapi-task.h`, function `OS_TaskDelay()`:
- [x] ONLY IN DOXY: OS_ERROR
- [x] OK: OS_SUCCESS

File `src/os/inc/osapi-task.h`, function `OS_TaskSetPriority()`:
- [x] ONLY IN DOXY: OS_ERROR
- [x] OK: OS_ERR_INVALID_ID
- [x] ONLY IN DOXY: OS_ERR_INVALID_PRIORITY
- [x] OK: OS_SUCCESS

File `src/os/inc/osapi-task.h`, function `OS_TaskGetIdByName()`:
- [x] OK: OS_ERR_NAME_NOT_FOUND
- [x] OK: OS_ERR_NAME_TOO_LONG
- [x] OK: OS_INVALID_POINTER
- [x] OK: OS_SUCCESS

File `src/os/inc/osapi-task.h`, function `OS_TaskGetInfo()`:
- [x] OK: OS_ERR_INVALID_ID
- [x] OK: OS_INVALID_POINTER
- [x] OK: OS_SUCCESS

File `src/os/inc/osapi-task.h`, function `OS_TaskFindIdBySystemData()`:
- [x] ONLY IN DOXY: OS_INVALID_POINTER
- [x] ONLY IN DOXY: OS_SUCCESS

**Additional context**
Originally noted in #331, split into separate work items. 

**Requester Info**

jphickey:
**Describe the contribution**
Ensures correlation between the unit-tests and documented return values for the OSAL task API.

Fixes #1021 

**Testing performed**
Build and run unit tests, re-run cross reference script to confirm

**Expected behavior changes**
All documented return values are tested or noted that they cannot be tested in functional test

**System(s) tested on**
Ubuntu

### DIFF
--- a/src/os/inc/osapi-task.h
+++ b/src/os/inc/osapi-task.h
@@ -92,11 +92,11 @@ typedef osal_task((*osal_task_entry)(void)); /**< @brief For task entry point */
  * the system heap.
  *
  * @param[out]  task_id will be set to the non-zero ID of the newly-created resource
- * @param[in]   task_name the name of the new resource to create
- * @param[in]   function_pointer the entry point of the new task
+ * @param[in]   task_name the name of the new resource to create @nonnull
+ * @param[in]   function_pointer the entry point of the new task @nonnull
  * @param[in]   stack_pointer pointer to the stack for the task, or NULL
  *              to allocate a stack from the system memory heap
- * @param[in]   stack_size the size of the stack
+ * @param[in]   stack_size the size of the stack @nonzero
  * @param[in]   priority initial priority of the new task
  * @param[in]   flags initial options for the new task
  *
@@ -105,10 +105,10 @@ typedef osal_task((*osal_task_entry)(void)); /**< @brief For task entry point */
  * @retval #OS_INVALID_POINTER if any of the necessary pointers are NULL
  * @retval #OS_ERR_INVALID_SIZE if the stack_size argument is zero
  * @retval #OS_ERR_NAME_TOO_LONG name length including null terminator greater than #OS_MAX_API_NAME
- * @retval #OS_ERR_INVALID_PRIORITY if the priority is bad
+ * @retval #OS_ERR_INVALID_PRIORITY if the priority is bad @covtest
  * @retval #OS_ERR_NO_FREE_IDS if there can be no more tasks created
  * @retval #OS_ERR_NAME_TAKEN if the name specified is already used by a task
- * @retval #OS_ERROR if an unspecified/other error occurs
+ * @retval #OS_ERROR if an unspecified/other error occurs @covtest
  */
 int32 OS_TaskCreate(osal_id_t *task_id, const char *task_name, osal_task_entry function_pointer,
                     osal_stackptr_t stack_pointer, size_t stack_size, osal_priority_t priority, uint32 flags);
@@ -125,7 +125,7 @@ int32 OS_TaskCreate(osal_id_t *task_id, const char *task_name, osal_task_entry f
  * @return Execution status, see @ref OSReturnCodes
  * @retval #OS_SUCCESS @copybrief OS_SUCCESS
  * @retval #OS_ERR_INVALID_ID if the ID given to it is invalid
- * @retval #OS_ERROR if the OS delete call fails
+ * @retval #OS_ERROR if the OS delete call fails @covtest
  */
 int32 OS_TaskDelete(osal_id_t task_id);
 
@@ -149,6 +149,7 @@ void OS_TaskExit(void);
  * @param[in]   function_pointer function to be called when task exits
  *
  * @return Execution status, see @ref OSReturnCodes
+ * @retval #OS_ERR_INVALID_ID if the calling context is not an OSAL task
  */
 int32 OS_TaskInstallDeleteHandler(osal_task_entry function_pointer);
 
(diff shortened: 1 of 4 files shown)


skliper (changes requested):
OS_TaskCreate - should test NULL stack address case (allocate from heap)
OS_TaskInstallDeleteHandler - check that delete handler actually gets called
OS_TaskDelay - add a check the delay actually delays? we depend on this working all over the place but I never see it actually checked
OS_TaskPriority - add a check that setting priority is actually respected
OS_TaskGetId - automate this check, pass in the pointer to the ID?
OS_TaskFindIdBySystemData - needs test

skliper:
I suggest we hold off on this one until @jphickey and I have a chance to discuss

jphickey:
Here's my take:

- OS_TaskSetPriority: No real easy way to check this in the current API, we would have to confine multiple tasks to the same CPU.  As long as tasks are allowed to float among SMP CPUs, a lower priority and a higher priority task can run at the same time.  Once we can assign affinity then the priority should be more verifiable.  (except when permissive mode is used....)
- OS_TaskFindIdBySystemData: As noted in the comment, there is no way to get the "system data" that this depends on.  This is only used by the PSP, where that info was acquired externally/outside of the OSAL API, to do the reverse lookup.  Hence why the test only checks the errors right now.
- OS_TaskGetId: This _is_ checking the returned value matches the expected ID.  I am not sure what is being requested on this one.
- OS_TaskDelay: I suppose we can add a call to OS_gettime before and after, then look at the difference

jphickey:
Commit 1870d32a adds a check for OS_TaskCreate with NULL stack (dynamically allocate), checks that the task delete handler was invoked, and confirms that OS_TaskDelay actually delayed by comparing with OS_GetLocalTime.

skliper:
@astrogeco @ejtimmon Since priority/scheduling testing is pretty much impossible to verify in a generic sense (without system/configuration knowledge), how about extending the OSAL example work to include a sample test that could confirm the MCP750 specific system setup (single core, whatever task scheduling is used)?  So this wouldn't be a built-in open-source OSAL unit test, but could be used to show things work as expected on the MCP750.  Would be great if this could eventually be extended to include a multi-cpu test, linux and RTEMS... but will still be system/configuration dependent.  I'll mark/report testing of priorities and scheduling as a project/distribution test responsibility and not confirmed generically using OSAL unit tests.

skliper (approved):

skliper:
Concur w/ all responses, fixes.  Good to go.  Note OS_TaskGetId comment was stale, so no longer applicable.

jphickey:
If we resurrect the tasking-example as has been discussed, that might be a good place to demonstrate priorities.  

I also thought we could check semaphore priorities.  That is, if we create a low priority task first, let it block, then create a high priority task, and have it block on the same sem.  If we then give the sem, only the high priority task should wake up.  If we give it a second time then the low priority task should wake up.  This would confirm priority-based semaphores which IIRC we are using, if the sem was configured as a FIFO then it would fail this test.

That test case could theoretically be done in the current tests but we would likely have to bypass it when OSAL_CONFIG_DEBUG_PERMISSIVE_MODE is enabled.

Also, the `OS_task_prop_t` structure has a "priority" member in it, which we can check.  However, this still gets set even if OSAL_CONFIG_DEBUG_PERMISSIVE_MODE is enabled.  Might want to modify this so it reflects the _real_ priority as programmed into the OS Scheduler, then we can use this field to gauge the behavior to expect from other calls.

jphickey:
Re-Pushed to correct commit message typo and squash commits where necessary

skliper:
> I also thought we could check semaphore priorities.... 

Good suggestion!  Could you submit a new issue?

> That test case could theoretically be done in the current tests but we would likely have to bypass it when OSAL_CONFIG_DEBUG_PERMISSIVE_MODE is enabled.

Isn't there a set of user level priorities that do still work in permissive?  Can we stay in that range?

> Also, the `OS_task_prop_t` structure has a "priority" member in it, which we can check. However, this still gets set even if OSAL_CONFIG_DEBUG_PERMISSIVE_MODE is enabled. Might want to modify this so it reflects the _real_ priority as programmed into the OS Scheduler, then we can use this field to gauge the behavior to expect from other calls.

Good suggestion/update!  Could you submit a new issue?

jphickey:
Submitted issue #1059 for above

PR #1049, opened 2021-05-24
Fix #1048, filter only whole words for keyword match

### LINKED ISSUE #1048: Stub generator script filtering all "extern" strings
**Describe the bug**
In order to parse the headers the "generate_stubs.pl" script needs to filter out "extern" keywords.  Problem is, it is filtering the "extern" string even if it appears inside a larger string, not just the keywords.

**To Reproduce**
View the generated stub for "OS_TimeBaseCreate".  The last parameter should be `external_sync` but it its appearing in the generated code as `al_sync` instead.

**Expected behavior**
Should appear in the generated code as `external_sync`, same as header.

**Code snips**
https://github.com/nasa/osal/blob/1d183e9a079036ff0c91f36eab9c1f29a69547f2/src/ut-stubs/osapi-timebase-stubs.c#L41

**System observed on:**
Ubuntu

**Reporter Info**

jphickey:
**Describe the contribution**
In the stub generator script, filter only whole words when ignoring keywords.  Do not match if the word appears as a substring inside a larger word.

Fixes #1048

**Testing performed**
Regenerate stubs, and build and run all unit tests.

**Expected behavior changes**
Corrects the argument name in OS_TimeBaseCreate stub (only occurrence discovered so far)

**System(s) tested on**
Ubuntu

### DIFF
--- a/src/ut-stubs/osapi-timebase-stubs.c
+++ b/src/ut-stubs/osapi-timebase-stubs.c
@@ -38,13 +38,13 @@ extern void UT_DefaultHandler_OS_TimeBaseGetInfo(void *, UT_EntryKey_t, const UT
  * Generated stub function for OS_TimeBaseCreate()
  * ----------------------------------------------------
  */
-int32 OS_TimeBaseCreate(osal_id_t *timebase_id, const char *timebase_name, OS_TimerSync_t al_sync)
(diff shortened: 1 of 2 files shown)


jphickey:
Note, it was brought to my attention that C++-style comments presented an issue for the script, they got incorrectly put into the output.  Updated this issue/PR to also include a fix for that.

skliper (approved):

PR #1070, opened 2021-06-09
Fix #1069, add detail design template

### LINKED ISSUE #1069: Provide file for cFE detail design
By default, when building cFE mission "detail design" document, the entire module directory is added to the doxygen input, which will recursively pick up all source files, but also everything else too.  

This can lead to conflicts with e.g. "mainpage", where OSAL provides its own mainpage for its API guide, but this should NOT be used for other documentation that includes OSAL, like the detail design doc.

OSAL should provide a detaildesign template file that indicates which sources should be used, rather than everything.

cFE itself could exclude certain parts in its templates, but that means cFE must have knowledge of OSAL directory tree.

**Additional context**
Related to nasa/cfe#1612

**Requester Info**

jphickey:
**Describe the contribution**
This file indicates which input dirs to use for the cFE detail design document.  It should be controlled with OSAL, as this
knows the detail of the directory tree.

Fixes #1069 

**Testing performed**
Build cFE detail design document, confirm correct/expected content, with no duplicate "mainpage"

**Expected behavior changes**
Detail design document has `src/os` and `src/bsp` content, but NOT the mainpage that is specific for the OSAL API guide.

**System(s) tested on**
Ubuntu

**Additional context**
Related to nasa/cfe#1612

### DIFF
--- a/doc/osal-detaildesign.doxyfile.in
+++ b/doc/osal-detaildesign.doxyfile.in
@@ -1,0 +1,7 @@
+# All of the OSAL FSW code relevant for a detail design document is under
+# src/os and src/bsp, everything else is test and examples/support
+# Note this will include ALL OS/BSP layers, it does not currently filter
+# based on which OS/BSP is actually in use.
+INPUT    += @osal_MISSION_DIR@/src/os
+INPUT    += @osal_MISSION_DIR@/src/bsp
+


skliper (approved):

skliper:
Suggestion - if it was named generically (remove osal-) then it might be easier to include from a generic module sense.  Although the script could just search based on any prefix so no biggie.

PR #1073, opened 2021-06-10
HOTFIX:2021-06-15, Add missing OS_Select param doc

astrogeco:
**Describe the contribution**

Fix usersguide doxygen warning for OS_SelectSingle and OS_SelectMultiple msecs input parameter.

**Testing performed**

**Expected behavior changes**
No doxygen warning

**System(s) tested on**

**Additional context**
See https://github.com/nasa/cFS/pull/265/checks?check_run_id=2798133902

**Third party code**
None

### DIFF
--- a/src/os/inc/osapi-select.h
+++ b/src/os/inc/osapi-select.h
@@ -92,6 +92,8 @@ typedef enum
  *
  * @param[in,out] ReadSet  Set of handles to check/wait to become readable
  * @param[in,out] WriteSet Set of handles to check/wait to become writable
+ * @param[in] msecs Indicates the timeout. Positive values will wait up to that many milliseconds. Zero will not wait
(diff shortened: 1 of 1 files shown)


skliper (approved):

jphickey (commented):
Looks good to me

skliper:
Is there background on why this is a hotfix?  It's been a while since any of this has changed, so why would it break now?  Was this missing from previous documentation builds?

jphickey:
I think the new/separate warning file perhaps made it more obvious that this was missing?

skliper:
Sounds like something might have been broken in CI then, maybe associated with the document building race... since the warning file was supposed to be getting checked for both users guides.

astrogeco:
Yeah, CI caught it that's why I had to fix it.

PR #1136, opened 2021-08-25
Fix #1135, add bitmask assert macros

### LINKED ISSUE #1135: Add UtAssert bit field check macros
Bit fields are a common method of storing multiple boolean values into a single data field.  UtAssert should provide macros that aid in testing code that uses this type of storage.

Add some new UtAssert macros to check bit fields.  These macros should include in the log the raw value of the bit field being checked (as a hex number) as well as the specific mask bits being checked for.  Two forms should be included, one which passes if the bit(s) are set, and one which passes if the bit(s) are not set.

This should be able to use the same underlying Generic assert helper function, just with a different macro wrapper around it that checks a bit mask.

May use the existing UtAssert_UINT32_EQ/NEQ, in either of these forms:

1. UtAssert_INT32_EQ(BitField, BitField | CHECK_BITS) 
2. UtAssert_INT32_EQ(BitField & CHECK_BITS, CHECK_BITS) 

Downsides: the resulting log is not quite as concise as it should be:
 - Its not clear that this is a bitmask check - as it is logged as a normal value check.
 - Only either the raw value (form 1) or the bitmask value (form 2) will be logged.  It will not log both pieces of info.
 - Values will be logged in decimal/base-10 by default which is not easy to read when examining bit flags - base-16 is preferable.

All of the above should be solvable by simply making a different wrapper macro around the existing `UtAssert_GenericUnsignedCompare` function.

**Additional context**
See nasa/cfe#1860 for a use case

**Requester Info**

jphickey:
**Describe the contribution**
Add a pair of macros that can confirm a value has bits set or does not have bits set.  By using bitmask-aware macros, the logged
information can include both the raw/actual value as well as the specific bits being tested.

Fixes #1135

**Testing performed**
Build and run all tests

**Expected behavior changes**
None here, just adds new macros

**System(s) tested on**
Ubuntu

**Additional context**
Tested by updating the new time clock state checks to use this macro.  Looks a little something like this in the log:

```
[BEGIN] 94 Test Clock
[ INFO] time_current_test.c:126:Testing: CFE_TIME_GetClockState, CFE_TIME_GetClockInfo
[ PASS] 94.001 time_current_test.c:145 - CFE_TIME_GetClockInfo() (0x33e0) &~ CFE_TIME_FLAG_CLKSET (0x8000)
[ PASS] 94.002 time_current_test.c:148 - CFE_TIME_GetClockInfo() (0x33e0) & CFE_TIME_FLAG_SRCINT (0x2000)
[ PASS] 94.003 time_current_test.c:149 - CFE_TIME_GetClockInfo() (0x33e0) & CFE_TIME_FLAG_SIGPRI (0x1000)
[ PASS] 94.004 time_current_test.c:150 - CFE_TIME_GetClockInfo() (0x33e0) &~ CFE_TIME_FLAG_REFERR (0x10)
[ PASS] 94.005 time_current_test.c:151 - CFE_TIME_GetClockInfo() (0x33e0) &~ CFE_TIME_FLAG_UNUSED (0xf)
[  END] 94 Test Clock           TOTAL::5     PASS::5     FAIL::0     MIR::0     TSF::0     TTF::0     WARN::0
```

### DIFF
--- a/ut_assert/inc/utassert.h
+++ b/ut_assert/inc/utassert.h
@@ -81,14 +81,16 @@ typedef enum
  */
 typedef enum
 {
-    UtAssert_Compare_NONE, /**< invalid/not used, always false */
-    UtAssert_Compare_EQ,   /**< actual equals reference value */
-    UtAssert_Compare_NEQ,  /**< actual does not non equal reference value */
-    UtAssert_Compare_LT,   /**< actual less than reference (exclusive) */
-    UtAssert_Compare_GT,   /**< actual greater than reference (exclusive)  */
-    UtAssert_Compare_LTEQ, /**< actual less than or equal to reference (inclusive) */
-    UtAssert_Compare_GTEQ, /**< actual greater than reference (inclusive) */
-    UtAssert_Compare_MAX   /**< placeholder, not used */
+    UtAssert_Compare_NONE,          /**< invalid/not used, always false */
+    UtAssert_Compare_EQ,            /**< actual equals reference value */
+    UtAssert_Compare_NEQ,           /**< actual does not non equal reference value */
+    UtAssert_Compare_LT,            /**< actual less than reference (exclusive) */
+    UtAssert_Compare_GT,            /**< actual greater than reference (exclusive)  */
+    UtAssert_Compare_LTEQ,          /**< actual less than or equal to reference (inclusive) */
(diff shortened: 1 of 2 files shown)


astrogeco:
[CCB:2021-09-01](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-01-2021) APPROVED

PR #1142, opened 2021-09-07
Fix #1141, add typecast to memchr call

### LINKED ISSUE #1141: Typecast memcchr in os-shared-common.h
**Describe the bug**
memchr is called and then assigned to a const char * without a cast.
This causes failures on certain compilers (llvm).

https://www.cplusplus.com/reference/cstring/memchr/
```
const void * memchr ( const void * ptr, int value, size_t num );
      void * memchr (       void * ptr, int value, size_t num );
```

In file `os-shared-common.h`
change 
```
    const char *end = memchr(s, 0, maxlen);
```
to 
```
    const char *end = (const char *)memchr(s, 0, maxlen);
```

**System observed on:**
 - Hardware
 - OS: [MAC]
 - Versions [OSAL Hash 7ebb463]
```
clang --version
Apple clang version 12.0.0 (clang-1200.0.32.29)
Target: x86_64-apple-darwin19.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

**Reporter Info**
Sam Price - GSFC

image

jphickey:
**Describe the contribution**
This function is documented as returning `void*`, and on some compilers this requires an explicit cast to `const char*` to avoid a warning.

Fixes #1141

**Testing performed**
Build and sanity check CFE

**Expected behavior changes**
None

**System(s) tested on**
Ubuntu

**Additional context**
Isolated from draft contribution from @thesamprice in #1140

### DIFF
--- a/src/os/shared/inc/os-shared-common.h
+++ b/src/os/shared/inc/os-shared-common.h
@@ -149,7 +149,7 @@ void OS_ApplicationShutdown_Impl(void);
  ------------------------------------------------------------------*/
 static inline size_t OS_strnlen(const char *s, size_t maxlen)
 {
-    const char *end = memchr(s, 0, maxlen);
+    const char *end = (const char *)memchr(s, 0, maxlen);
     if (end != NULL)
     {
         /* actual length of string is difference */


jphickey:
After further consideration, I think the reason this warning got tripped is because the function in question is an inline, and in the proposed code change it was possibly included from a C++ source file.  A void conversion like this is valid in C but in C++ requires a cast.

Nonetheless, this makes the headers C++ compliant so still a valid thing to add.

skliper (approved):

astrogeco:
Do we want to do a "co-authored by" credit?

thesamprice:
> Do we want to do a "co-authored by" credit?

Doesn't matter to me. Don't make more work than necessary.

astrogeco:
[CCB:2021-09-08](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-08-2021) APPROVED

PR #1169, opened 2021-09-28
Fix #1166, recognize ifdef __cplusplus

### LINKED ISSUE #1166: UT assert stub generator bypasses functions marked `extern "C"`
**Describe the bug**
The `generate_stubs.pl` script ignores all contents inside `{}` blocks.  Unfortunately, for header files which are made C++-aware, it is typical to wrap C function calls in an `extern "C"` block which means it gets ignored by the tool.  It does not understand the `#ifdef __cplusplus` surrounding it.

**To Reproduce**
Attempt to generate stubs against a header file which includes an extern "C" block.  The prototypes inside the block are not processed.

**Expected behavior**
Should process prototypes inside the extern "C" block.

**System observed on:**
Ubuntu.

**Additional context**
Prototypes were expected to all occur at the top-most scope but the `extern "C"` keyword for C++ changes this.  Really, the file needs to be run through the C (not C++) preprocessor first, so the `#ifdef` here has the intended effect.

**Reporter Info**

jphickey:
**Describe the contribution**
Add feature to generate_stubs.pl input phase to recognize and skip over `#ifdef __cplusplus` blocks.  This makes it so it
can be used on C headers which are C++-enabled too.

The logic should be fairly easy to extend to other `#if` conditionals as needed.

Fixes #1166

**Testing performed**
Generate stubs for a C header that has C++-aware extern wrappers in it.

**Expected behavior changes**
Stubs for prototypes inside C++ extern "C" block are generated successfully.

**System(s) tested on**
Ubuntu

**Additional context**
Tempting to attempt to use gcc proprocessor here (-E) to get full evaluation of all preprocessor logic, but this would depend on the target headers and may introduce more problems than it solves.  So in the interim this just adds a local check that specifically looks for `#ifdef __cplusplus` and filters out everything inside this block.

### DIFF
--- a/ut_assert/scripts/generate_stubs.pl
+++ b/ut_assert/scripts/generate_stubs.pl
@@ -110,6 +110,7 @@ foreach my $hdr (@hdrlist)
     my $file = "";
     my $file_boilerplate;
     my $file_variadic;
+    my @ifdef_level = (1);
 
     # All header files start with some legal boilerplate comments
     # Take the first one and save it, so it can be put into the output.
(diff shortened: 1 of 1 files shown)


jphickey:
Note I did not test this with anything nested deeply/complexly - for example it currently does not handle `#elif` blocks.

astrogeco:
https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-29-2021 APPROVED

astrogeco:
Merged in 4a2caddae72ea022f4a9a58f243ad78b0e0b9ca2

Issue #780, opened 2021-02-09
File created without restricting permissions in unit test tools


skliper:
**Is your feature request related to a problem? Please describe.**
CodelQL security warnings against following code:
https://github.com/nasa/osal/blob/09a2c5e793639e508174a2446989274acf36f25e/ut_assert/src/uttools.c#L60
https://github.com/nasa/osal/blob/09a2c5e793639e508174a2446989274acf36f25e/ut_assert/src/uttools.c#L102-L124

**Describe the solution you'd like**
https://github.com/nasa/elf2cfetbl/blob/6d1d12ea2d4ab7bcc7044f8e4c60a268d7a1f640/elf2cfetbl.c#L1455-L1464

**Describe alternatives you've considered**
None

**Additional context**
CodeQL analysis results

**Requester Info**
Jacob Hageman - NASA/GSFC

PR #1181, opened 2021-10-13
Fix #1175, Use fstat and fchmod for TOCTOU Bug

### LINKED ISSUE #1175: TOCTOU Bug for chmod
**Describe the bug**
Calling function chmod that uses Filename after a check function. This can cause a time-of-check, time-of-use race condition.

**Expected behavior**
Use fchmod as seen in OS_FileChmod_Impl.

**Code snips**
https://github.com/nasa/osal/blob/42af0f73c952b4e4c7fbc06c97d354a1430f4ac3/ut_assert/src/uttools.c#L62-L69

https://github.com/nasa/osal/blob/42af0f73c952b4e4c7fbc06c97d354a1430f4ac3/ut_assert/src/uttools.c#L111-L117

**System observed on:**
Coverity: https://scan.coverity.com/projects/arielsadamsnasa-cfs-jsf-rules?tab=overview

**Reporter Info**
Ariel Adams, ASRC Federal

ArielSAdamsNASA:
**Describe the contribution**
Fixes  #1175 

**Testing performed**
GitHub Actions

**Expected behavior changes**
Resolve TOCTOU bug by using fstat and fchmod instead of stat and chmod. 

Ariel Adams, ASRC Federal

### DIFF
--- a/ut_assert/src/uttools.c
+++ b/ut_assert/src/uttools.c
@@ -57,14 +57,15 @@ typedef struct
 bool UtMem2BinFile(const void *Memory, const char *Filename, uint32 Length)
 {
     FILE *      fp;
+    int         fd;
     struct stat dststat;
 
     if ((fp = fopen(Filename, "w")))
     {
-        if (stat(Filename, &dststat) == 0)
+        fd = fileno(fp);
+        if (fstat(fd, &dststat) == 0)
         {
-            chmod(Filename, dststat.st_mode & ~(S_IRGRP | S_IWGRP | S_IXGRP | S_IROTH | S_IWOTH | S_IXOTH));
-            stat(Filename, &dststat);
+            fchmod(fd, dststat.st_mode & ~(S_IRGRP | S_IWGRP | S_IXGRP | S_IROTH | S_IWOTH | S_IXOTH));
         }
 
         fwrite(Memory, Length, 1, fp);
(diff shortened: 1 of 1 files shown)


skliper (changes requested):
fstat and fchmod take the file descriptor, not the file name as an input, as in `fstat(fd, &dststat)`.  See https://linux.die.net/man/2/fstat.  Also, the second fstat doesn't seem to do anything  useful, so it could probably be removed.

skliper (approved):

skliper:
@dzbaker @dmknutsen ef44a74 is failing.  I don't think this one was ready to merge... causing failures in all other repo unit tests.

jphickey:
Just came here because I was seeing some strange build failures - confirm this needs to be reverted - it is definitely breaking stuff.

Problem is that it uses POSIX APIs (fileno, fstat) in code that is supposed to be plain C99.

jphickey:
NOTE: I'd actually propose reverting the whole thing - that is, not just the use of fstat/fchmod but also the use of stat/chmod that came before it.  

It looks like the genesis of all this was a codeQL warning back here: #780

HOWEVER - context is important, these are coverage tests and not running on a shared server.  C99 does not have any concept of "file permissions" - that's a POSIX thing.  This code is supposed to be portable and should adhere to strict C99, nothing more.  That means we can't change file permissions - on a UNIX-type system this will be controlled by the user's "umask" setting.  So permission _is_ already controlled - codeQL was generating a false positive warning here IMO in saying that permissions were not set.  They were, just not directly in the source code.

skliper:
I agree 100% w/ @jphickey.  We've snowballed in complexity trying to squash warnings that aren't really applicable.

PR #1206, opened 2022-01-12
Fix #1204, Search global and local symbol tables

### LINKED ISSUE #1204: `OS_SymbolLookup` should also support local symbol lookup
Getting symbol addresses are an integral part of MM and MD.  For systems that load modules with local scope this could still be easily supported by looping through the modules w/o requiring app changes.

Currently it just looks for global or static symbols:
https://github.com/nasa/osal/blob/4cc6dbb5019d0589d5ce52e3755a0b7a012ade3c/src/os/shared/src/osapi-module.c#L352-L388

Possibly just use OS_ForEach and loop through modules within OS_SymbolLookup, but on systems where local symbol table lookup just calls the global version that would waste cycles if a symbol doesn't exist since it would end up searching the global table for every module.

Consider a slight refactor... a generic OS_SymbolLookup_Impl could search just global on systems with only global, global and local on systems that support it.  Really doesn't seem to be a use case for OS_GlobalSymbolLookup_Impl which only searches the global table... possibly remove?

None others.

**Additional context**
See MM and MD use of symbol lookup, helpful to get from global or local scope.

**Requester Info**
Jacob Hageman - NASA/GSFC

skliper:
**Describe the contribution**
- Fix #1204 
- Refactors symbol table searching to include both local and global symbol tables for POSIX
- Renamed global search to generic since there  isn't currently a use case for global only search

**Testing performed**
DRAFT - not complete yet

**Expected behavior changes**
Improved portability, recovers prior behavior and will find symbols even if loaded as local only.

**System(s) tested on**
 - Hardware: i5/Docker
 - OS: Ubuntu 18.04
 - Versions: Bundle main

**Additional context**
None

**Third party code**
None

Jacob Hageman - NASA/GSFC


skliper:
Currently in draft for preliminary review, still need to update unit tests.

astrogeco:
[CCB:2022-01-19](https://github.com/nasa/cFS/wiki/CCB:-2022.01.19) APPROVED
- Had to be implemented at the implementation level
- Adds test to find in both global and local symbols

jphickey (approved):

PR #1209, opened 2022-01-18
Fix #1208, typesafe definition of osal_id_t

### LINKED ISSUE #1208: Type-safe implementation for osal_id_t
The `osal_id_t` type is currently just a plain `uint32` to make it backward-compatible.  The problem is that a plain number does not convey the purpose/intent of this ID value.  Users can still do things like addition, subtraction, or interchange with other numbers, which are all invalid for a number being used as an ID.

As long as users are using the `osal_id_t` type and the associated checks/conversions also provided by OSAL (e.g. OS_ObjectIdDefined, etc) then this should be transparent.

Make `osal_id_t` type-safe

**Additional context**
This was the intent all along with creating a separate `osal_id_t` typedef and using it across the CFE and PSP.
It does depend on having all the dependent code using the typedef, so probably want to stage this in using OSAL_OMIT_DEPRECATED flag or similar, so we can find the things that are still broken in this regard.

**Requester Info**

jphickey:
**Describe the contribution**
Modifies the osal_id_t typedef to be a non-integer value.  The intent is to catch cases where it inappropriately being used as an integer value.

This is transparent so long as the osal_id_t typedef and provided check and conversion routines are used.

Fixes #1208 

**Testing performed**
Build and sanity check CFE and OSAL with and without OSAL_OMIT_DEPRECATED flag

**Expected behavior changes**
When OSAL_OMIT_DEPRECATED is enabled (opt-in) this will catch misuse of OSAL IDs as integers, or failure to use correct `osal_id_t` typedef to hold the value.

**System(s) tested on**
Ubuntu 21.10

**Additional context**
OSAL and CFE should all be compliant, but CFS apps probably are not yet compliant with this.  Hence why it is "opt-in" via OSAL_OMIT_DEPRECATED flag, like other similar changes done in the past.

### DIFF
--- a/src/os/inc/common_types.h
+++ b/src/os/inc/common_types.h
@@ -91,10 +91,24 @@ extern "C"
     typedef size_t    cpusize;
     typedef ptrdiff_t cpudiff;
 
+#ifdef OSAL_OMIT_DEPRECATED
     /**
      * A type to be used for OSAL resource identifiers.
+     * This is a type-safe ID, and cannot be implicitly converted to an integer.
+     * Use the provided inline functions in osapi-idmap.h to interpret ID values.
      */
-    typedef uint32_t osal_id_t;
+    typedef struct
+    {
+        uint32_t v;
+    } osal_id_t;
+#else
+
+/**
+ * A type to be used for OSAL resource identifiers.
+ * This typedef is backward compatible with the IDs from older versions of OSAL
+ */
+typedef uint32 osal_id_t;
+#endif
 
     /**
      * A type used to represent a number of blocks or buffers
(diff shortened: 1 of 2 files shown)


jphickey:
There is one exception to CFE compliance here, see nasa/cfe#2024
The fix for the CFE issue should be merged first.

skliper (approved):

astrogeco:
[CCB:2022-01-19](https://github.com/nasa/cFS/wiki/CCB:-2022.01.19) APPROVED

astrogeco:
Missed merging it to IC:Caelum+dev2, added this PR to https://github.com/nasa/cFS/pull/414 and it passess all checks so merging straight to main.

PR #1223, opened 2022-02-23
Fix #1222, protect if OS_FDGetInfo called on socket

### LINKED ISSUE #1222: Segfault when calling OS_FDGetInfo on datagram socket ID
**Describe the bug**
Sockets opened with OS_SocketOpen get assigned an OSAL ID which is of the same type/classification as file IDs and shares the same global table.  However, these IDs do not have a name associated with them, and thus the `name_entry` field of the global structure is NULL for these objects.

The `OS_FDGetInfo` will validate this type of ID (because it is a file type) but will attempt to copy the name unconditionally, which results in a segfault because the pointer is NULL.

**To Reproduce**
1. Acquire a datagram socket ID using `OS_SocketOpen()`
2. Invoke `OS_FDGetInfo()` on that socket
3. Get Segfault in `strncpy()`

**Expected behavior**
Should not segfault, rather the returned name should be empty.

**Code snips**
Offending call to strncpy is here:
https://github.com/nasa/osal/blob/b0958583748eedd776622f0a873c769b7c554d48/src/os/shared/src/osapi-file.c#L535

**System observed on:**
Ubuntu 

**Additional context**
This code just needs to account for the name_entry being (validly) NULL on some types of entries.

**Reporter Info**
Originally found by @dmccomas in OSK implementation

jphickey:
**Describe the contribution**
Datagram sockets do not have a name_entry, it is set NULL.  If the user calls OS_FDGetInfo() on this type of ID, it will
pass the first test, so this needs to be checked for non-NULL.

Fixes #1222

**Testing performed**
Build and sanity check OSAL
Run all tests
Confirm coverage test is exercising the new check

**Expected behavior changes**
Calling OS_FDGetInfo on a datagram socket returns an empty string in the Path field, it does not segfault.

**System(s) tested on**
Ubuntu

### DIFF
--- a/src/os/shared/src/osapi-file.c
+++ b/src/os/shared/src/osapi-file.c
@@ -532,7 +532,10 @@ int32 OS_FDGetInfo(osal_id_t filedes, OS_file_prop_t *fd_prop)
     {
         record = OS_OBJECT_TABLE_GET(OS_global_stream_table, token);
 
-        strncpy(fd_prop->Path, record->name_entry, sizeof(fd_prop->Path) - 1);
+        if (record->name_entry != NULL)
+        {
+            strncpy(fd_prop->Path, record->name_entry, sizeof(fd_prop->Path) - 1);
+        }
         fd_prop->User    = record->creator;
         fd_prop->IsValid = true;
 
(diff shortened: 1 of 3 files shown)


skliper (approved):
Strange the `name` wasn't used before.  Approved.

jphickey:
> Strange the `name` wasn't used before. Approved.

Yeah, I was surprised by that too.  Every existing caller was passing `"ABC"` as the name, so the fact that the `name` parameter wasn't actually used wasn't noticed, until I tried to add a call where it was set NULL....

dmccomas:
Changes correct the issue I was having.

PR #1238, opened 2022-03-25
Fix #1237, avoid calling memchr() with unknown size buffer

### LINKED ISSUE #1237: `UtAssert_StringBufCompare` SEGV when -1 for size passed in (on CentOS 7)
**Describe the bug**
`memchr(x, x, -1)` returns NULL on CentOS 7 (at least in one case), which within `UtAssert_StringBufCompare` causes the `memcmp` to  use the max size_t which causes a SEGV.

https://github.com/nasa/osal/blob/fafb045e7a5311e19affe0ddab724f1604e739f4/ut_assert/src/utassert.c#L459-L513

**To Reproduce**
Build and run this test case (others also use this pattern):
https://github.com/nasa/cFE/blob/a39b0a65fb2724692469ff492484b523fb4fa7e6/modules/config/ut-coverage/test_cfe_config.c#L138-L146

**Expected behavior**
No SEGV

**Code snips**
See above

**System observed on:**
 - Hardware: unknown
 - OS: CentOS 7
 - Versions: Bundle main

**Additional context**
[Add any other context about the problem here.](https://stackoverflow.com/questions/47315902/is-it-legal-to-call-memchr-with-a-too-long-length-if-you-know-the-character-wil)

**Reporter Info**
Jacob Hageman - NASA/GSFC

jphickey:
**Describe the contribution**
In some systems, passing a large size value to memchr() causes it to return NULL, even if the char is guaranteed to be found within the actual valid buffer memory.

This modifies the string buffer comparison function to actively check for this sentinel value and use "strlen()" instead.

Fixes #1237

**Testing performed**
Build and sanity check, run all tests on both Ubuntu (latest version) and CentOS 7 VM

**Expected behavior changes**
Tests pass in both systems (no SEGV on CentOS 7)

**System(s) tested on**
Ubuntu 21.10
CentOS 7


skliper (approved):

jphickey:
Note - format check issue appears to be a problem that was already in the build (not caused by this PR)

PR #1250, opened 2022-04-25
FIX #1248, add os-impl-no-select.c for RTEMS

### LINKED ISSUE #1248: Add os-impl-no-select.c for operating systems that do not have select (RTEMS 6 with network off)
RTEMS 6.x removes the legacy network stack by default, and in doing so, removes the "select" call. Now, when the OSAL OSAL_CONFIG_INCLUDE_NETWORK option is set to FALSE, the OSAL/cFS build fails because it still expects a select call.

I have implemented a src/os/portable/os-impl-no-select.c file along with the corresponding changes to the CMakeLists.txt file that uses os-impl-no-select.c when the INCLUDE_NETWORK option is FALSE, otherwise it uses os-impl-bsd-select.c as usual.

**Requester Info**
Alan Cudmore / NASA GSFC / Code 582.0

acudmore:

**Describe the contribution**

- Fix #1248 

Implements os-impl-no-select.c for RTEMS or any other OS that has a configuration without select.
Currently only used for RTEMS when the network stack is disabled in the OSAL. On RTEMS the select call only works with network file descriptors.
In current RTEMS 4.11 and 5 builds, when the OSAL_CONFIG_INCLUDE_NETWORK option is false, the select call will still be included even though it cannot be used. The does not fail because the select call is still available in the RTEMS libraries.

If you build RTEMS 4.11 or 5 without the default network stack, the select call will not be included in the RTEMS libraries, causing the OSAL build to fail. In addition, RTEMS 6 removes the legacy network stack, so a default OS build will cause the OSAL build to fail.

By selecting os-impl-no-select.c when the OSAL_CONFIG_INCLUDE_NETWORK option is false, the select call will be omitted, and it will build for RTEMS with and without the network stack.

**Testing performed**
This was tested with RTEMS 4.11 and RTEMS 5 using the pc-rtems CI platform.
I ran all tests on 4.11 and 5 with OSAL_CONFIG_INCLUDE_NETWORK set to true and false. All tests pass, but I did notice that the OSAL module unload test did not pass on RTEMS 4.11 - I don't think this test is related.
This is also being used on LEON3/RTEMS 5.1 for GTOSat.

**Expected behavior changes**
When building the cFS with the OSAL_CONFIG_INCLUDE_NETWORK to false for RTEMS only (currently), the OS_Select* calls will return OS_ERR_NOT_IMPLEMENTED. This is probably safer, since the underlying select call should only be used with network sockets/file descriptors anyway.

**System(s) tested on**
 - Build system: Docker containers for RTEMS 4.11 and 5 cFS CI tests
 - Target OS/cFS Platform: RTEMS 4.11/x86 with cFS pc-rtems platform, and RTEMS 5/x86 with cFS pc-rtems platform
 - This patch has also been in use for GTOSat on a LEON3/RTEMS 5.1 platform for a year.

Alan Cudmore, NASA GSFC/582.0

### DIFF
--- a/src/os/portable/os-impl-no-select.c
+++ b/src/os/portable/os-impl-no-select.c
@@ -1,0 +1,78 @@
+/************************************************************************
+ * NASA Docket No. GSC-18,719-1, and identified as “core Flight System: Bootes”
+ *
+ * Copyright (c) 2020 United States Government as represented by the
+ * Administrator of the National Aeronautics and Space Administration.
+ * All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain
+ * a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ************************************************************************/
+
+/**
+ * \file   os-impl-no-select.c
+ * \author joseph.p.hickey@nasa.gov
+ *
+ * Purpose: All functions return OS_ERR_NOT_IMPLEMENTED.
+ * This is used when network functionality is disabled by config.
+ */
+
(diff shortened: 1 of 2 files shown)


astrogeco:
[CCB:2022-04-27](https://github.com/nasa/cFS/wiki/CCB:-2022.04.27) APPROVED

PR #1257, opened 2022-05-20
Fix #1256, add rtems 6 support

### LINKED ISSUE #1256: Add OSAL support for RTEMS 6
RTEMS 6 is currently under development in the main branch of the RTEMS git repository. There are a few minor changes that have to be addressed in the OSAL code to support RTEMS 6.

Modify the OSAL RTEMS implementation and OSAL pc-rtems BSP to support RTEMS 6 while still supporting RTEMS 4.11 and RTEMS 5.
This is done with approximately 6 minor changes. In the existing code there are conditional blocks selecting either RTEMS 5 and defaulting to RTEMS 4.11. For example:

```
ifdef _RTEMS_5_
   RTEMS 5 specific statement or define
else
   RTEMS 4.11 specific statement or define
endif
```

To support RTEMS 6, these blocks can be flipped as so:

```
ifdef OS_RTEMS_4_DEPRECATED
   RTEMS 4.11 specific statement or define
else
   RTEMS 5 and 6 (and hopefully future versions)
endif
```

By flipping these blocks, RTEMS 5 and 6 can be supported in the else block.

Alternatives include not supporting RTEMS 6, or just supporting RTEMS 6, neither of which are ideal. This solution will continue to support RTEMS 4.11, RTEMS 5, and add support for RTEMS 6.

**Additional context**
These changes depend on a number of other changes:
https://github.com/nasa/cFE/issues/2096
https://github.com/nasa/PSP/issues/338
https://github.com/nasa/osal/issues/1248
I will not submit a pull request for this fix until the above changes are all in the main branch of the cFS bundle.

**Requester Info**
Alan Cudmore, NASA/GSFC Code 582.0

acudmore:

**Describe the contribution**

- Fix #1256

These changes add support for RTEMS 6 in the RTEMS OSAL and pc-rtems OSAL BSP. The changes consist of modifications to the conditional compilation blocks and a couple of new conditional compilation blocks in the pc-rtems OSAL BSP startup code to account for an include and API that was deprecated in RTEMS 6.
In the RTEMS OSAL, the existing conditional compilation blocks were flipped from:

```
if RTEMS 5
   RTEMS 5 specific statement(s)
else
   RTEMS 4.11 specific statement(s)
endif
```

to:

```
if RTEMS 4.11
   RTEMS 4.11 specific statement(s)
else
   RTEMS 5 and above specific statement(s)
endif
```

By flipping the blocks as described above, it should work with future versions of RTEMS provided the API does not change.

There is also a change in the os-impl-loader.c file to solve a warning with GCC 12. GCC 12 is now used by RTEMS 6.

**Testing performed**
All tests were run on the QEMU RTEMS 4.11, QEMU RTEMS 5, and the new QEMU RTEMS 6 containers.

**Expected behavior changes**
There should be no impact to the behavior. All tests passed on RTEMS 4.11, 5, and 6.

**System(s) tested on**
QEMU x86 with pc-rtems RTEMS BSP and cFS pc-rtems platform. In addition these changes are being used on a generic-rtems platform that will be submitted in a future PR.

**Additional context**
This is the last set of code changes in the first round of RTEMS 6 updates described here:
https://github.com/nasa/cFS/discussions/468

**Third party code**
None

Alan Cudmore, NASA/GSFC Code 582.0

### DIFF
--- a/src/bsp/pc-rtems/build_options.cmake
+++ b/src/bsp/pc-rtems/build_options.cmake
@@ -5,13 +5,13 @@
 ##########################################################################
 
 # Link the RTEMS BSP with the "rtemscpu" system library
+target_link_libraries(osal_bsp
+      rtemscpu
+)
+
+# Add the networking library for RTEMS 6+
 if(CMAKE_SYSTEM_VERSION GREATER 5)
     target_link_libraries(osal_bsp
-         rtemscpu
          networking
-    )
-else()
-    target_link_libraries(osal_bsp
-          rtemscpu
     )
 endif(CMAKE_SYSTEM_VERSION GREATER 5)
--- a/src/bsp/pc-rtems/build_options.cmake
+++ b/src/bsp/pc-rtems/build_options.cmake
@@ -5,6 +5,13 @@
 ##########################################################################
 
 # Link the RTEMS BSP with the "rtemscpu" system library
-target_link_libraries(osal_bsp
-    rtemscpu
-)
+if(CMAKE_SYSTEM_VERSION GREATER 5)
+    target_link_libraries(osal_bsp
+         rtemscpu
+         networking
+    )
+else()
+    target_link_libraries(osal_bsp
+          rtemscpu
+    )
+endif(CMAKE_SYSTEM_VERSION GREATER 5)
--- a/src/bsp/pc-rtems/src/bsp_start.c
+++ b/src/bsp/pc-rtems/src/bsp_start.c
@@ -35,7 +35,6 @@
 #include <ctype.h>
 #include <bsp.h>
 #include <rtems.h>
-#include <rtems/mkrootfs.h>
 #include <rtems/bdbuf.h>
 #include <rtems/blkdev.h>
 #include <rtems/diskdevs.h>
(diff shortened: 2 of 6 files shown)


src/bsp/pc-rtems/build_options.cmake
```diff
@@ -5,6 +5,13 @@
...
+if(CMAKE_SYSTEM_VERSION GREATER 5)
+    target_link_libraries(osal_bsp
+         rtemscpu
+         networking
+    )
+else()
+    target_link_libraries(osal_bsp
+          rtemscpu
+    )
+endif(CMAKE_SYSTEM_VERSION GREATER 5)
```
jphickey, line 17: Just a minor/pedantic comment that "target_link_libraries" is defined by CMake to append libraries to the link list, it does not replace or change existing link libraries.  Therefore this could be simplified, where "rtemscpu" is always added first (as it was) and then just add the RTEMS 6 specific library like so:

```
if(CMAKE_SYSTEM_VERSION GREATER 5)
    target_link_libraries(osal_bsp networking)
endif()
```

This would also simplify things a bit if someday we want to build an RTEMS 6 image without networking - because one would still need rtemscpu in that case too.

jphickey (approved):

src/bsp/pc-rtems/build_options.cmake
acudmore: Good suggestion, I'll make that change.

PR #1278, opened 2022-08-10
Fix #1252, condition variable API for OSAL

### LINKED ISSUE #1252: Add abstraction for condition variables
POSIX condition variables can be useful for implementing worker thread pools, allowing worker task(s) to pend for work and dispatcher/manager tasks to unblock those workers once a job becomes available.  

Abstraction of something like POSIX condition variables via OSAL APIs.

Some of the worker pool concepts may be implemented with simple semaphores, but those aren't ideal for some use cases:

- a counting sem needs a 1:1 relationship between gives and takes, so it is not ideal for cases where work can be "batched" - either where a single input can unblock multiple worker jobs, or multiple inputs can be handled by a single worker job.
- a binary sem helps the second case by capping the semaphore count at 1, such that multiple "gives" on the dispatch side only cause a single worker run.  But it does not allow for a dispatcher to unblock multiple workers, nor can it act on other conditions than a simple boolean.
 
Note that in the POSIX OSAL the binary semaphore is, in fact, implemented using condition variables underneath - where the condition is just a single boolean value.  But only this limited use-case is exposed/available to applications.  

As a backward-compatible compromise of sorts - it may be possible to build on the current binary sem concept but allow the application to register its own "condition test" routine beyond the simple boolean.  

**Additional context**
VxWorks support is not fully clear, I think VxWorks 7 has the underlying APIs to implement this, but 6.x might not.  

**Requester Info**

jphickey:

**Describe the contribution**
Implementation of condition variable API for OSAL

Currently implemented only for POSIX, RTEMS can be added with future commit.

**Testing performed**
Coverage and functional test of condvar API is included

**Expected behavior changes**
Adds condition variable API

**System(s) tested on**
Ubuntu 22.04

**Additional context**
The BP/DTN implementation (https://github.com/nasa/bplib) makes use of condition variables to signal between worker tasks.  In order to support running this library natively, OSAL would need to provide this set of functionality.


dzbaker:
CCB 8 September 2022: @jphickey will write issues for extending to additional platforms.

PR #1280, opened 2022-08-20
Fix #1279, rtems allow variable length queue msgs.

### LINKED ISSUE #1279: RTEMS osal queue mixed sizes
**Describe the bug**
Im using queues to send variable length messages.
RTEMS queue get returns an error if the message queue returned size is not the max buffer size.
https://github.com/nasa/osal/blob/main/src/os/rtems/src/os-impl-queues.c#L235

This doesn't match the posix implementation.
https://github.com/nasa/osal/blob/main/src/os/posix/src/os-impl-queues.c#L194

The above logic in `shared/src/osapi-queue.c` uses `OS_QUEUE_INVALID_SIZE` If the queue size is greater than the buffer provided.
```c
        if (size < queue->max_size)
        {
            /*
            ** The buffer that the user is passing in is potentially too small
            */
            *size_copied = 0;
            return_code  = OS_QUEUE_INVALID_SIZE;
        }
```

**To Reproduce**
Visual inspection.

**Expected behavior**
Follow the posix logic.

**Code snips**
See above

**System observed on:**
MIcroblaze

**Additional context**

**Reporter Info**
Sam Price

thesamprice:

* [X] I reviewed the [Contributing Guide](https://github.com/nasa/osal/blob/main/CONTRIBUTING.md).
* [X] I signed and emailed the appropriate [Contributor License Agreement](https://github.com/nasa/cFS/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla) to GSFC-SoftwareRelease@mail.nasa.gov and copied cfs-program@lists.nasa.gov.
NA as Gov employee? NA as removing code not adding?

**Describe the contribution**
A clear and concise description of what the contribution is.
Drops check on size != rtems_size to allow variable length queue messages.
Matches the posix queue implementation.

**Testing performed**
1. Build steps '...'
Built and ran a non cfs system using posix.
built and ran the same cfs system using rtems, this fix allowed system to run.

1. Execution steps '...'
Ran on RTEMS on microblaze.

**Expected behavior changes**
 - API Change: RTEMS queues changed to match posix queues for variable length messages 

**System(s) tested on**
RTEMS / Microblaze

**Additional context**

**Third party code**
N/A

NASA GSFC

### DIFF
--- a/src/os/rtems/src/os-impl-queues.c
+++ b/src/os/rtems/src/os-impl-queues.c
@@ -223,8 +223,7 @@ int32 OS_QueueGet_Impl(const OS_object_token_t *token, void *data, size_t size, 
     }
 
     /*
-    ** Check the size of the message.  If a valid message was
-    ** obtained, indicate success.
+    ** If a valid message was obtained fill in queue size, otherwise fill in zero on error
     */
     if (status == RTEMS_SUCCESSFUL)
     {
--- a/src/os/rtems/src/os-impl-queues.c
+++ b/src/os/rtems/src/os-impl-queues.c
@@ -229,11 +229,6 @@ int32 OS_QueueGet_Impl(const OS_object_token_t *token, void *data, size_t size, 
     if (status == RTEMS_SUCCESSFUL)
     {
         *size_copied = rtems_size;
-        if (rtems_size != size)
-        {
-            /* Success, but the size was wrong */
-            return_code = OS_QUEUE_INVALID_SIZE;
-        }
     }
     else
     {


thesamprice:
Im not sure how to fix the commit message.
On my branch its correct. See [link](https://github.com/nasa/osal/commit/57184695a6315c91fc8324925bd3d8d1bd759905)
Somehow the pull request is changing the commit message.

skliper:
@thesamprice - it was the pull request title (lowercase `f` on `fix`) that was causing the job to fail.  We haven't figured out yet how to get the job to pick up the corrected title to be able to rerun and have it pass.

skliper (dismissed):
Maybe make the size verification optional?  Just skip the check if passed in size is zero.

thesamprice:
@skliper 
The size passed in is the maximum buffer size that rtems is allowed to write in, passing in zero would break functionality.

skliper:
Maybe instead of just deleting the check (and checking for success twice), it could be updated to match the pattern in POSX/VxWorks.  Unless someone comes back with a reason why the check could be required for RTEMS I don't have any issues with making it consistent.

thesamprice:
If someone is relying on this check in rtems, their code will break when going to vxworks / posix.

I would recommend  doing the following only if the check is needed.
- moving the check up into the shared area.
- and adding some sort of settings flag when its initialized 
OS_FIXED_SIZE_QUEUE
- only doing the check if OS_FIXED_SIZE_QUEUE is set.

This would keep the existing functionality common across all platforms.

The functionality should be documented also.

I use the queues to route messages to udp/tcp so this rtems check is breaking for me only on rtems.

Ill go ahead and fork osal, and make the change so our system can work on rtems.

dzbaker:
CCB 25 Aug 2022: @dzbaker will pick up from @skliper

thesamprice:
Closing pull request, this will be replaced by the posix mimic version.

PR #1291, opened 2022-09-23
Fix #1289, cmake script modernization

### LINKED ISSUE #1289: Clean up remaining deprecated/obsolete features used by the CMake scripts
In recent years the CMake build has been adopting the current best-practices whenever feasible to do so, and moving away from features that have been marked as obsolete or deprecated.

- Use interface properties to propagate compiler flags, definitions, etc to dependencies, rather than setting "magic variables" that need to be known by the dependency
- Prefer `target_include_directories` as opposed to `include_directories` at directory scope
- Prefer `target_compile_definitions` as opposed to `add_definitions` at directory scope
- Use of `target_link_libraries` should include the scope keyword (PUBLIC/PRIVATE) rather than the "bare" signature (also noted by CFE user in nasa/cfe#2141)

In particular - the following specific areas should be modernized:

- Public API location should be an interface library: https://github.com/nasa/osal/blob/38559d447fdeb1a85f8bb09938df30d7a50918a4/CMakeLists.txt#L89
- Management of impl-specific flags can also be done via interface library rather than manual propagation here: https://github.com/nasa/osal/blob/38559d447fdeb1a85f8bb09938df30d7a50918a4/CMakeLists.txt#L154-L163
- Use scope keyword on link libraries: https://github.com/nasa/osal/blob/38559d447fdeb1a85f8bb09938df30d7a50918a4/CMakeLists.txt#L282
- Implement UT coverage interface library as alternative to "flag variables" exported here: https://github.com/nasa/osal/blob/38559d447fdeb1a85f8bb09938df30d7a50918a4/CMakeLists.txt#L354-L355

**Additional context**
It is not currently possible to propagate the UT coverage flags if OSAL and the application are in separate builds.

Using the target-scope features and interface libraries is of particular benefit to making a proper OSAL "package" that can be compiled and linked separately from the application (see #1284)

**Requester Info**

jphickey:

**Describe the contribution**
Use target properties to define interfaces and compiler definitions rather than referencing global variables or using "known" paths in other modules.  This better aligns with current practices and creates a more robust build environment that is less dependent on specific path names existing in a given module.

This adds an "osal_public_api" interface target that contains the paths to the public API headers as its INTERFACE_INCLUDE_DIRECTORIES and any required compiler definitions as its INTERFACE_COMPILE_DEFINITIONS property.  Applications should use this rather than referring to the "${OSAL_SOURCE_DIR}/src/os/inc" include path directly.

Fixes #1289 

**Testing performed**
Build and confirm OSAL successfully builds on various platforms including Ubuntu 22.04 (new-ish CMake) and 18.04 (old-ish CMake).

**Expected behavior changes**
No impact to runtime behavior (no actual source code changes here)

Build script makes more use of CMake "INTERFACE" libraries to communicate compiler options, include paths, and definitions to dependencies, moving away from using global variables to perform this task.  Notably, this permits the targets to be imported into another build more easily.

**System(s) tested on**
Ubuntu 18.04 and 22.04 (native, newer and older CMake)
RTEMS 4.11 and 5.0

**Additional context**
All changes should be backward compatible, in that the "magic variables" are still being honored as the current version of the CMake script had previously implemented:

- OSAL_API_INCLUDE_DIRECTORIES (input)
- OSAL_USER_C_FLAGS (input)
-  UT_COVERAGE_COMPILE_FLAGS (output)
-  UT_COVERAGE_LINK_FLAGS (output)

The preferred method of configuring these items going forward would be to get/set the corresponding properties on the `osal_public_api` interface target for include directories and compile options, and use the `ut_coverage_compile` and `ut_coverage_link` interface targets for coverage testing flags.  Defining both methods for now allows for transition.  

If the OSAL_OMIT_DEPRECATED flag is set, then the UT_COVERAGE flags will not be exported.


chillfig:
Approved. Confirmation that testing on VxWorks is preferable before merging this pull request.

jphickey:
Update - I did confirm the build using ppc-vxworks6.9 and all was fine, no issues observed.
Should be OK to merge!

PR #1313, opened 2022-10-12
Fix #1284, add export targets and package script

### LINKED ISSUE #1284: Standalone OSAL build should export a package for use by other builds
OSAL is not standalone software that is typically installed on a PC, but rather it is always a component that is intended to be used within a larger software application.

Currently the CMake build does work in a standalone fashion (i.e. one can invoke `cmake </path/to/osal>` and build the software) and the software can be compiled but the result isn't really usable for inclusion in a larger app.  That is, there is no easy way for another (separate) application that needs to use OSAL to reference the headers and static library.

CMake has a method for referencing a dependency through its import/export and packaging functionality described here: https://cmake.org/cmake/help/latest/guide/importing-exporting/index.html

**Additional context**
Note this is _not_ necessary when building cFS - in that mode, OSAL is just included as a subdirectory within the larger build, and this should still be supported as an option for other (non-CFS) builds as well.  That is, the user can simply call `add_subdirectory(/path/to/osal)` from their application build, and link to the provided `osal` library as a regular target.  The export target is only useful in cases where OSAL is built as a standalone software application.

**Requester Info**

jphickey:

**Describe the contribution**
Adds optional "install" commands to CMake script, which is useful when building OSAL as a standalone software package.  The public API and static libraries will be installed to the specified system location, and can then be used to compile and link external OSAL applications without the need for the original OSAL source or build trees.

Note this also installs the "osconfig.h" file as this does affect the binary formats of some items (i.e. size of items using OS_MAX_API_NAME and other similar structure members).  The external application must be compiled using the same osconfig.h after installation.

Fixes #1284

**Testing performed**
Build OSAL with both CFE and standalone and run all tests (existing build)

Then run "make install" on standalone build to stage headers and library binaries to a build tree.  Tested an external application to using `find_package(NasaOsal)` and building against the OSAL headers and binaries provided by that exported package.

**Expected behavior changes**
Exports a CMake package containing targets that can be used by OTHER builds without necessarily having the full OSAL source tree integrated into that project.

**System(s) tested on**
Ubuntu 22.04

**Additional context**
This is necessary when developing standalone (non-CFE) software that still uses OSAL to provide portability to different systems and/or does any unit testing (the exported packages includes UT assert).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,15 @@
 #                  Normally this setting is not needed to be configured as it is
 #                  inferred from the BSP type.
 #
+# OSAL_INSTALL_LIBRARIES : Boolean, enables "install" of targets listed above such 
+#                  that libraries and public API header files are copied into the system
+#                  location specified by CMAKE_INSTALL_PREFIX.  This set of headers
+#                  and link libraries can then be used to compile other applications
+#                  separately from OSAL.  Default is "ON" when OSAL is being compiled
+#                  standalone (i.e. cmake called directly on the OSAL source dir).
+#                  Default is "OFF" when OSAL is included via 'add_subdirectory' in 
+#                  a parent application project such as CFE/CFS.
+#
 # ENABLE_UNIT_TESTS : Boolean, enables build of the unit tests (coverage and functional)
 #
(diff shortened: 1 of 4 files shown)


dzbaker:
CCB 13 October 2022: Approved pending workflow re-run.

PR #1352, opened 2023-01-19
Fix #1344, UT assert stub return values

### LINKED ISSUE #1344: Add better support for pointer return values in UT assert
Currently, UT assert only as built-in support for 32-bit integer return values, in the form of `UT_SetDefaultReturnValue` and `UT_SetDeferredRetcode`.   The majority of CFE/CFS APIs have a 32-bit integer return value, so this has been good enough.

However, there are also a fair number of CFE functions that return pointer values, and these `int32` APIs do not currently work for pointer returns.  Any stub that returns a pointer must have its own custom handler function to help set the return value, and so support is spotty.  Every function that has a pointer return may have a slightly different way to control the return value, and a different range of things that can be done by the test to modify its behavior.

Add two new APIs to UT assert:
- `UT_SetDefaultReturnObject`
- `UT_SetDeferredReturnObject`

These would work like the current integer functions, respectively, but accept pointers with a size instead of an integer.  This type of signature could support ALL types of return values, including 8-bit, 16-bit, 64-bit, and potentially even larger structures (even though code should not output large structures via return value, it could be done).

A halfway point could be to just add generic reusable handlers that return a pointer, and could be registered to any function.  But that would not be as clean, and would still need a separate function for each type.

**Additional context**
This can be added without breaking any currently-existing tests or stubs, because its only a new type of return record.  It would not change anything about the existing logic, just extend it.

**Requester Info**

jphickey:

**Describe the contribution**
Improves default support for return values of sizes other than 32 bits. Previously this required a custom handler for any value that was not a 32 bit integer.  With this update any integer value can be used, and a value translation will occur if the size is different.  Thus boolean values, 8,16, and 64-bit ints can be configured through the same basic API without needing to add a handler.
Fixes #1344

**Testing performed**
Build and run all tests, including CFS app tests, to confirm all still run as expected.

**Expected behavior changes**
No change to _existing_ behavior - all currently-implemented test cases should continue to behave as they have in the past, whether they used a 32-bit integer return or if they had a custom hook/handler.

Going forward, it should no longer be required to implement a custom handler for the sole purpose of adjusting the size of the return value in case it is not a 32-bit integer.  As a fallback default, any return value that is unset by a test will be memset to 0.  So a pointer stub return will get `NULL`, a boolean will get `FALSE`, and any numeric value of any time will be zero.  

The `UT_SetDeferredRetcode` and `UT_SetDefaultReturnValue` functions now accept a `UT_IntReturn_t` value instead of `int32`.  This typedef is defined using the C99 `ptrdiff_t` type, which will be 64 bits in size on modern hardware.  When passing by value from an int32, this should be transparent, but it is large enough to also allow test cases to pass in pointer values here without truncating.

**System(s) tested on**
Debian 

**Additional context**
API to set return values is still numeric in nature.  This adds some framework to allow fully-typed values and abstract objects bigger than `ptrdiff_t` values to be returned, but there is no API yet - that can be added in a future change.

### DIFF
--- a/ut_assert/inc/utgenstub.h
+++ b/ut_assert/inc/utgenstub.h
@@ -40,7 +40,7 @@
  * \param ReturnType The type of return value
  */
 #define UT_GenStub_SetupReturnBuffer(FuncName, ReturnType) \
-    UT_Stub_RegisterReturnType(UT_KEY(FuncName), sizeof(ReturnType))
+    UT_Stub_RegisterReturnType(UT_KEY(FuncName), sizeof(ReturnType), #ReturnType)
 
 /**
  * Helper macro to get the return value from the handler
@@ -51,7 +51,7 @@
  * \param ReturnType The type of return value
  */
 #define UT_GenStub_GetReturnValue(FuncName, ReturnType) \
-    (*(ReturnType *)UT_Stub_GetReturnValuePtr(UT_KEY(FuncName), sizeof(ReturnType)))
+    (*(ReturnType *)UT_Stub_GetReturnValuePtr(UT_KEY(FuncName), sizeof(ReturnType), #ReturnType))
 
 /**
  * Helper macro to add a local parameter to the current context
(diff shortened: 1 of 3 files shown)


dzbaker:
CCB 19 January 2022: Will review after next release.

chillfig (approved):

Issue #185, opened 2019-09-30
OS_SymbolLookup Function Prototype Variable Names Do Not Match Implementation


skliper:
The OS_SymbolLookup function prototype (in osapi-os-loader.h) defines the prototype to be:

int32 OS_SymbolLookup (cpuaddr *symbol_address, const char *symbol_name );

While the implementations of this function declares the function with variable names that do not match the prototype:

int32 OS_SymbolLookup( cpuaddr *SymbolAddress, const char *SymbolName )

This could result in compiler errors/warnings on certain platforms.  It is recommended to update the prototype variable names to match the implementations.

skliper:
_Imported from trac issue 162.  Created by sstrege on 2016-03-10T12:21:00, last modified: 2019-04-23T15:45:34_

skliper:
_Trac comment by cdknight on 2018-07-24 13:57:19:_

CCB determined this is not worth working.

PR #1354, opened 2023-01-23
Fix #1353, Align mismatched function prototype/implem. parameter names

### LINKED ISSUE #1353: Scrub for mismatched function prototype/implementation parameter names
**Checklist**
* [x] I reviewed the [Contributing Guide](https://github.com/nasa/cFS/blob/main/CONTRIBUTING.md).
* [x] I performed a cursory search to see if the bug report is relevant, not redundant, nor in conflict with other tickets.

**Describe the bug**
Noticed a couple of mismatched parameter names between function prototypes and their implementations. A quick scrub of OSAL turned up 13 cases in total.

**Expected behavior**
While not a risk to behavior, these mismatched parameter names may cause compiler errors in the future on some systems, and can be confusing for maintenance and application programming.

**Code snips**
Cases identified:
image

**Reporter Info**
Avi Weiss @thnkslprpt

thnkslprpt:
**Checklist**

**Describe the contribution**
Fixes #1353 
Similar to https://github.com/nasa/cFE/pull/1551/files from cFE

May as well clean these up now to avoid any potential future compiler warnings and ease maintainability/readability.

Note: one of these cases was noticed in https://github.com/nasa/osal/issues/185, but closed as 'won't fix' at the time.

**Testing performed**
GitHub CI actions (incl. Build + Run, Unit Tests etc.) all passing successfully.

**Expected behavior changes**
Function prototypes and their implementations should (ideally) always have matching parameter names.

**Contributor Info**
Avi Weiss @thnkslprpt


src/os/inc/osapi-file.h
```diff
@@ -382,7 +382,7 @@ int32 OS_remove(const char *path);
  * @retval #OS_FS_ERR_PATH_TOO_LONG if the paths given are too long to be stored locally
  * @retval #OS_FS_ERR_NAME_TOO_LONG if the new name is too long to be stored locally
  */
-int32 OS_rename(const char *old_filename, const char *new_filename);
+int32 OS_rename(const char *old, const char *new);
```
jphickey, line 385: We should strive to avoid changing the name in the header/declaration, but rather prefer to update the implementation to match the declaration whenever possible.

Note that in this particular case `new` is a reserved keyword in C++ but not reserved in C.  There are some C++ users that would get tripped up if the name were changed to `new`.

jphickey (changes requested):
I concur with the general desire to make the names of the parameters consistent between the declaration and definition, but in this case the names in the headers (API) are probably the preferred names for these values.  For example, in the "timebase" module, the IDs should all be "timebase_id" rather than "timer_id" (the latter probably a remnant of cut-and-paste from the timer API).  Somewhere along the way, the headers got updated appropriately, but the implementation wasn't.  We should just update the implementation, not revert the header.

As a side note, cppcheck should be finding and reporting differences in these parameter names, but does not seem to be flagging these items.  This might point to a problem with our cppcheck workflow not covering everything it is supposed to.

src/os/inc/osapi-file.h
thnkslprpt: Good point re. the reserved keyword.
Will update this and the rest of the implementations to match the prototypes.

PR #1428, opened 2023-11-17
Fix #1427, Add default build/install scripts

### LINKED ISSUE #1427: Add default build/install scripts for regular and test builds.
Add default build/install scripts for regular and test builds.
The purpose is to build directly from unmodified scripts.

Assume the parent folder of osal can have `build-osal` and `local` folder trees.
Use relative paths `../build-osal` and `../local` to build and install.

Suggested scripts: `mk-osal-no-test.sh` - default build and `mk-osal-test.sh` - build with test.

gskenned:
**Description**
Fixes issue 1427. Added two default build/install scripts.
`mk-osal-no-test.sh` - Builds OSAL and installs in ../local
`mk-osal-test.sh` - Builds OSAL, runs OSAL tests, and installs in ../local

**Testing performed**
1. Ran both build scripts successfully.
1. All OSAL tests in the build passed.

**Expected behavior changes**
The default build scripts provide example build commands using relative paths.

**System(s) tested on**
 - Hardware: AWS
 - OS: Ubuntu 20.04 LTS
 - Versions: OSAL 5.0.99

Grafton S. Kennedy, III - Vantage Systems, Inc.

### DIFF
--- a/mk-osal-no-test.sh
+++ b/mk-osal-no-test.sh
@@ -1,0 +1,8 @@
+mkdir -p ../build_osal
+cd ../build_osal
+cmake -DOSAL_SYSTEM_BSPTYPE=generic-linux \
+-DCMAKE_INSTALL_PREFIX=../local \
+../osal
+make
+make install
+
--- a/mk-osal-test.sh
+++ b/mk-osal-test.sh
@@ -1,0 +1,10 @@
+mkdir -p ../build_osal_test
+cd ../build_osal_test
+cmake -DENABLE_UNIT_TESTS=true \
+-DOSAL_SYSTEM_BSPTYPE=generic-linux \
+-DOSAL_CONFIG_DEBUG_PERMISSIVE_MODE=TRUE \
+-DCMAKE_INSTALL_PREFIX=../local \
+../osal 
+make
+make test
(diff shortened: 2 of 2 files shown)


gskenned:
Changed "Fix nasa#1427" to "Fix #1427" in PR and commit comment per format checker.

jphickey:
I'm not sure if we should have these as `.sh` scripts.  Although I think its good information to capture, I get nervous if someone says "just run this mystery script" ... 

For one, it is missing the `!#` at the start, but also it doesn't handle any errors, e.g. if the "mkdir" or "cd" fails, then you'll end up running cmake in the wrong dir.  Or something.

Instead of giving a script that is directly in script form -- my recommendation would be to add this to the README, and let the user put it (e.g. copy/paste) into their own script file.  This is safer, as it doesn't make assumptions about the user's system, and the onus is on the user to make sure its correct for their system before they go and run it.

gskenned:
Thank you for your review [comments](https://github.com/nasa/osal/pull/1428#issuecomment-1819893542), Joe. I agree with all of them. This PR is ready to close. Mostly my intention was to document scripts that I shared with another team member.

PR #1450, opened 2024-02-14
Fix #1449, POSIX implementation honors stack pointer

### LINKED ISSUE #1449: POSIX implementation does not adhere to user-specified stack pointer
The `OS_TaskCreate()` function accepts a stack pointer argument.  Per the API, if this is not NULL, it is supposed to be used as the stack pointer for the newly-created task.  VxWorks and RTEMS implement this, but POSIX ignores this argument.  Currently, the POSIX implementation will allocate a stack buffer, regardless of what was passed in.

Should adhere to the passed-in stack pointer argument.

**Additional context**
POSIX does define a function for this - `pthread_attr_setstack()`.
Related issue in CFE: nasa/cfe#2516 ... when validating that fix, noticed that the newly-added tests fail on POSIX, because it does not adhere to the stack pointer.

**Requester Info**

jphickey:

**Describe the contribution**
When the OS_TaskCreate() function is called with a non-NULL stack pointer, the POSIX implementation now should use this pointer as intended.  Previously it was always allocating a new stack buffer, ignoring the passed-in pointer.

NOTE: When using this feature, the caller must use caution to make sure the buffer being passed in is large enough to use as a task stack.  Particularly when running under a desktop Linux/Glibc environment, the infrastructure puts a considerably large object on the stack.  For example, in the osal-core-tests, the stack size was too small.

Fixes #1449

**Testing performed**
Build and run all tests.  Added tests that ensure the stack pointer is being honored as expected.

**Expected behavior changes**
User-supplied stack buffer will be used on POSIX.

**System(s) tested on**
Debian

**Additional context**
This could have negative consequences if the user was assuming RTOS-like behavior (e.g. no thread local storage) and was using very small stack sizes.  If the supplied stack size is less than 16k, (or `PTHREAD_STACK_MIN`) then the call will now fail, whereas previously it would quietly be made bigger than requested.

### DIFF
--- a/src/examples/tasking-example/tasking-example.c
+++ b/src/examples/tasking-example/tasking-example.c
@@ -30,7 +30,7 @@
 /* Task 1 */
 
 #define TASK_1_ID         1
-#define TASK_1_STACK_SIZE 1024
+#define TASK_1_STACK_SIZE 4096
 #define TASK_1_PRIORITY   101
 
 uint32 task_1_stack[TASK_1_STACK_SIZE];
@@ -40,7 +40,7 @@ void task_1(void);
 /* Task 2 */
 
 #define TASK_2_ID         2
-#define TASK_2_STACK_SIZE 1024
+#define TASK_2_STACK_SIZE 4096
 #define TASK_2_PRIORITY   102
 
 uint32 task_2_stack[TASK_2_STACK_SIZE];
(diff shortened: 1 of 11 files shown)


jphickey:
Ready for [re-]review.  Checking further into detail of the RTEMS implementation -- the needed API was only added in the latest development branch (6.x) and is not available on either RTEMS 4.11 nor 5.x.  Furthermore, the documentation warns that this is only intended for use with low-level device drivers that have special requirements, and general tasks should _not_ force a stack pointer, as the stack buffers have special requirements for size and alignment.

Therefore, with the consideration that we will likely also deprecate/discourage users from specifying a stack buffer due to similar issues on other platforms, the best solution for now is to simply skip this functional test on RTEMS.

PR #1465, opened 2024-06-13
Fix #1458, Adds JSC 2.1 Static Analysis comments

### LINKED ISSUE #1458: Static analysis issues JSC 2.1
**Describe the bug**
Handful of static analysis issues. This report records the findings from the static analysis to facilitate their tracking.

**To Reproduce**
N/A. These issues were communicated by JSC.

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:w="urn:schemas-microsoft-com:office:word"
xmlns:m="http://schemas.microsoft.com/office/2004/12/omml"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Word.Document>
<meta name=Generator content="Microsoft Word 15">
<meta name=Originator content="Microsoft Word 15">
<link rel=File-List
href="file:////Users/jdfiguer/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">

<link rel=themeData
href="file:////Users/jdfiguer/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_themedata.thmx">
<link rel=colorSchemeMapping
href="file:////Users/jdfiguer/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_colorschememapping.xml">

</head>

<body lang=EN-US link=blue vlink="#96607D" style='tab-interval:.5in;word-wrap:
break-word'>

Source Folder | Source Filename | Source Line | Source Object | Disposition
-- | -- | -- | -- | --
Integer Overflow |   |   |   |  
\osal\src\os\inc | osapi-clock.h | 218 | tm | Will Not Fix. Overflow is not considered a concern because   tm would need to be over 29,227 years in microseconds.
Memset Bound Checking |   |   |   |  
\osal\src\os\shared\src\osapi-condvar.c | 69 | memset |   | Will Not Fix. Using memset as sizeof(OS_condvar_table)   ensures correct array size.
\osal\src\os\shared\src\osapi-condvar.c | 294 | memset |   | Will Not Fix. Using memset as sizeof(OS_condvar_prop_t)   ensures correct array size.
Replace strlen with strnlen   \osal\src\os\shared\src\osapi-condvar.c | 301 | strncpy |   | Will Fix. osal PR# 1465
Unchecked return value on Function/Truncation of Data Risk |   |   |   |  
\osal\src\os\shared\src\osapi-errors.c | 113 | snprintf |   | Will Not Fix. No need to check snprintf return;   OS_STATUS_STRING_LENGTH (12) is ample for all status values

</body>

</html>

**Expected behavior**
Need to resolve these. Should resolve and/or disposition the higher ranked ones at minimum.

**Code snips**
N/A

**System observed on:**
N/A

**Additional context**
N/A

**Reporter Info**
Imported from JSC 2.1 static analysis

chillfig:

**Describe the contribution**
- Fixes #1458, Adds dispositions to line of code in question from static analysis stating why it's safe to keep as-is.

**Testing performed**
Manual inspection

**Expected behavior changes**
N/A

**System(s) tested on**
N/A

**Additional context**
N/A

**Third party code**
N/A

Justin Figueroa, Vantage Systems


src/os/inc/osapi-clock.h
```diff
@@ -215,6 +215,7 @@ static inline int64 OS_TimeGetTotalMicroseconds(OS_time_t tm)
  */
 static inline OS_time_t OS_TimeFromTotalMicroseconds(int64 tm)
 {
+    /* SAD: Overflow is unlikely, requiring an input equivalent to over 29.23 years in microseconds. */
```
chillfig, line 218: CCB 06/13/2024: Change "Overflow is unlikely..." to "This is not considered a concern because...etc"
Also verify 29.23 years.

jphickey (commented):
Definitely some mixed feelings about this.  I thought we had concurred that because an overflow would only occur if the interval was over 29000 years, that it was OK as-is?

Seems like we are just shifting risk here -- although the tool might no longer complain about this, we haven't made the code any more correct.  That is, if you pass in an interval that is beyond the representable range, then returning `INT64_MAX` is not more (or less) correct than returning some unspecified/platform-defined value.  Both results would be equally incorrect.

In order to justify this change, all callers of this API would have to check for the `INT64_MAX` result, and (theoretically?) take some alternate action.  But that is not feasible - all these tests are not free, they take time and CPU power and increase the code size and they need to be unit-tested and functional-tested and increase our future maintenance load.  All for something that only occurs in the highly unlikely event that a user needs to represent a time interval greater than 29000 years, and probably won't be handled correctly anyway?

In summary - I prefer the previous approach of simply documenting the limits of what `OS_time_t` can represent, and instructions on how to adjust this (note that the user can change the time tick to represent a larger range or better precision, whatever is more important)

PR #1480, opened 2024-10-06
Fix #1440, Split up BinSemGetInfo() to avoid partial success returns

### LINKED ISSUE #1440: SemGetInfo partial success cases
`OS_BinSemGetInfo` and `OS_CountSemGetInfo` fill in the following structures:
https://github.com/nasa/osal/blob/1a8823f2dc93da2bcb3696c92c7aee4f512ed4f7/src/os/inc/osapi-binsem.h#L38-L44
https://github.com/nasa/osal/blob/1a8823f2dc93da2bcb3696c92c7aee4f512ed4f7/src/os/inc/osapi-countsem.h#L32-L37

Where name and creator may be valid but not all implementations fill in value (ex. RTEMS):
https://github.com/nasa/osal/blob/1a8823f2dc93da2bcb3696c92c7aee4f512ed4f7/src/os/rtems/src/os-impl-binsem.c#L275-L279

Note the following PR changes the RTEMS return to `OS_ERR_NOT_IMPLEMENTED`
- #1438

Consider splitting the API into 2 or 3.  GetValue could return value or unimplemented error.  GetName/GetCreator or similar could return the other two.

None

**Additional context**
Partial success API's are not user friendly and should be avoided.

**Requester Info**
Jacob Hageman - NASA/GSFC

thnkslprpt:
**Checklist**

**Describe the contribution**
- Fixes #1440 
  - Splits the `BinSemGetInfo()` API into 3, allowing clear success/failure (or not implemented) for each.

**Testing performed**
GitHub CI actions all passing successfully (incl. Build + Run, Unit/Functional Tests etc.).
Tested locally to confirm no loss of coverage compared with the current main branch, and all lines/branches/functions in the new version are covered.

**Expected behavior changes**
Removes this source of 'partial success' which can be confusing and also increases complexity for dealing with the return for users calling into this API.

**System(s) tested on**
Debian 12 using the current main branch of cFS bundle.

**Contributor Info**
Avi Weiss &nbsp; @thnkslprpt

### DIFF
--- a/src/os/inc/osapi-binsem.h
+++ b/src/os/inc/osapi-binsem.h
@@ -178,6 +178,8 @@ int32 OS_BinSemDelete(osal_id_t sem_id);
  */
 int32 OS_BinSemGetIdByName(osal_id_t *sem_id, const char *sem_name);
 
+#ifdef OSAL_OMIT_DEPRECATED
+#else
 /*-------------------------------------------------------------------------------------*/
 /**
  * @brief Fill a property object buffer with details regarding the resource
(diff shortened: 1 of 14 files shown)


thnkslprpt:
Pretty simple change-set, although the change is 'breaking' I guess, so has to wait for a major release?

Also I think there is a single use of `OS_BinSemGetInfo()` in cFE maybe - so that small change will need to go along with this if it is merged.

skliper:
> Pretty simple change-set, although the change is 'breaking' I guess, so has to wait for a major release?
> 
> Note - there is a [single reference](https://github.com/nasa/cFE/blob/78b8fd7d4509a3a122787f7eddb2fb21c05402b5/modules/es/ut-coverage/es_UT.c#L1903) to `BinSemGetInfo()` in cFE which will need to be updated along with this if it is merged.

I recommend just deprecating the old API vs fully removing, then it doesn't have to be a major release (at least per the old rules I used).  See other uses of `OSAL_OMIT_DEPRECATED`.

thnkslprpt:
@skliper I'm not sure if I did the deprecation guards correctly, but do you know if it's possible or have the debug and release builds pass the tests given that the deprecated elements are only removed in the debug workflow?
https://github.com/nasa/osal/blob/827016689c1acba664843f18c4081be4dc174743/.github/workflows/standalone-build.yml#L60

Is it trying to run the same set of tests, but with and without the deprecated code? (that doesn't seem to make sense right)

(I'm assuming we only upkeep tests for the non-deprecated code...)

skliper:
> @skliper I'm not sure if I did the deprecation guards correctly...

I'm not really following your full question, but I recommend only wrapping the old API in the omit deprecated ifndef not the new stuff.  Then hopefully the tests would pass?

thnkslprpt:
> > @skliper I'm not sure if I did the deprecation guards correctly...
> 
> I'm not really following your full question, but I recommend only wrapping the old API in the omit deprecated ifndef not the new stuff. Then hopefully the tests would pass?

OK all good now - it wasn't building for me locally but that might have been a symptom of my own build command. Tests passing with a small addition as per the above comment.

PR #1569, opened 2026-08-21
Fix nasa#1568, task affinity support vxrtp

### LINKED ISSUE #1568: Add VxWorks RTP task affinity support
Add task affinity support for the vxworks-rtp operating system to support SMP operations.

arodarrow:

**Describe the contribution**
Adding SMP API to OSAL for VxWorks RTP

**Testing performed**
Built with VxWorks RTP on the vxsim. Looked at event messages at startup and used get and set affinity cmds to verify expected values.

**Expected behavior changes**
Able to change task affinity and get core num information

**System(s) tested on**
- VxSIM(on a linux VM)
- OS: VxWorks rtp

Adrian Rodriguez NASA GSFC

### DIFF
--- a/src/os/vxworks-rtp/src/os-impl-pthread-affinity.c
+++ b/src/os/vxworks-rtp/src/os-impl-pthread-affinity.c
@@ -19,50 +19,42 @@
 /**
  * \file
  *
- * This file Contains all of the api calls for manipulating files
- * in a file system / C library that implements the POSIX-style file API
+ * OSAL Task Affinity Implementation for VxWorks RTP
+ * This bridges the POSIX OSAL task structure with native VxWorks CPU Affinity APIs
  */
 
 /****************************************************************************************
                                     INCLUDE FILES
  ***************************************************************************************/
 
-/*
- * Inclusions Defined by OSAL layer.
- *
- * This must include whatever is required to get the prototypes of these functions:
- *
- *   open()
- *   stat()
- *   chmod()
- *   remove()
- *   rename()
- */
-
 #include <stdio.h>
 #include <string.h>
 #include <errno.h>
 #include <time.h>
 
+/* POSIX thread support (needed for the OSAL internal records) */
+#include <pthread.h>
+
+/* Native VxWorks headers required for Task Affinity in RTP */
+#include <vxWorks.h>
+#include <cpuset.h>
+#include <taskLib.h>
+#include <vxCpuLib.h>
+#include <errnoLib.h>
+
 #include "os-impl-taskaffinity.h"
 #include "osapi-task-affinity.h"
-
 #include "os-impl-tasks.h"
 #include "os-shared-file.h"
 #include "os-shared-idmap.h"
 
 /****************************************************************************************
-                                     DEFINES
+                                 Task Affinity API
  ***************************************************************************************/
 
-/****************************************************************************************
-                                 Named File API
- ***************************************************************************************/
 /*
  * ----------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


src/os/vxworks-rtp/src/os-impl-tasks.c
```diff
@@ -0,0 +1,924 @@
...
+   Name: OS_PthreadEntry
+
+   Purpose: A Simple pthread-compatible entry point that calls the real task function
+
+   returns: NULL
+
+    NOTES: This wrapper function is only used locally by OS_TaskCreate below
+
+---------------------------------------------------------------------------------------*/
+static void *OS_PthreadTaskEntry(void *arg)
```
dmolock-NASA, line 134: Instead of duplicating the entire file you could move the OS_PthreadTaskEntry() routine to a separate file and have a standard POSIX version and a VxWorks-RTP version that stores the VxWorks task id.

src/os/vxworks-rtp/src/os-impl-pthread-affinity.c
```diff
@@ -0,0 +1,144 @@
...
+    cpuset_t                        local_cpuset;
+    OS_impl_task_internal_record_t *impl = OS_OBJECT_TABLE_GET(OS_impl_task_table, *token);
+    uint32                          i;
+    int32                           return_value;
+
+    /* Check if the thread has actually started and recorded its ID yet */
+    if (impl->vxid == 0 || impl->vxid == (TASK_ID)ERROR)
+    {
+        return OS_ERROR;
+    }
```
jphickey, line 124: This check is not needed - likely impossible to reach except in unit tests.

"Impl" routines in OSAL should be lightweight translation units between the generic OSAL data structures and the OS-specific data structures.  They should not need to do general error checking, just specific error checking associated with the OS call.

If the "vxid" value in the impl structure is not valid, then VxWorks will reject it during `taskCpuAffinitySet()`.  We do not need to (double-) check it here.  It should never be invalid in practice.

src/os/vxworks-rtp/src/os-impl-pthread-affinity.c
```diff
@@ -0,0 +1,144 @@
...
+        for (i = 0; (i < OS_MAX_CPUS) && (i < OS_TaskAffinityGetCoresConfigured()); i++)
+        {
+            if (CPUSET_ISSET(local_cpuset, i))
+            {
+                OS_CPUSET_SET(i, cpuset);
+            }
+        }
+        return OS_SUCCESS;
+    }
+    return OS_ERROR;
```
jphickey, line 143: Can we avoid multiple return statements here?

Might be just a style thing -- but if we must `return` from inside a conditional, then it should be an error check that does no additional work.  This way the typical/nominal flow is in the main function part, not in the conditional.  (basically just reverse the logic in the `if` check to make it an error check instead of a success check).

Also -- is there any additional info that can be printed in an OS_DEBUG statement when the call to `taskCpuAffinityGet()` does fail?  Does VxWorks have something akin to an errno that we can get and print?  It is always good to include error detail whenever possible, not just a generic OS_ERROR unspecified fault that will be impossible to track down the cause if it ever happens.

jphickey (changes requested):
Minor suggestions for cleanup, overall looks OK

src/os/vxworks-rtp/src/os-impl-tasks.c
arodarrow: @jphickey do you think I should do that approach?

PR #1573, opened 2026-08-29
fix(posix): make mutex robust, handle EOWNERDEAD

philphauler:
Fixes #2433

Set PTHREAD_MUTEX_ROBUST and recover via pthread_mutex_consistent on EOWNERDEAD. Prevents deadlock when owning thread cancelled during shutdown.

AI Used: y - reviewed.

### DIFF
--- a/src/os/posix/src/os-impl-mutex.c
+++ b/src/os/posix/src/os-impl-mutex.c
@@ -1,4 +1,4 @@
-/************************************************************************
+﻿/************************************************************************
  * NASA Docket No. GSC-19,200-1, and identified as "cFS Draco"
  *
  * Copyright (c) 2023 United States Government as represented by the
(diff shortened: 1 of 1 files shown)


joelsherrill:
The robust APIs are not supported by RTEMS or required by the FACE Technical Standard. This code needs to account for it not being universally available.

Obviously, someone's free to submit the code required to support this in RTEMS.  :)

dzbaker:
@philphauler Thank you for your contribution. Please resolve the workflow failures.

sylvesterkaczmarek (commented):
I think the feature guard is ineffective on glibc/Linux: `PTHREAD_MUTEX_ROBUST` is an enum constant, not a preprocessor macro, so `#ifdef PTHREAD_MUTEX_ROBUST` evaluates false and the mutex never becomes robust. That leaves the original deadlock behavior intact on Linux. Could this be gated by an actual platform/feature-test macro (or CMake capability check) and covered by a Linux test that kills the owner and observes `EOWNERDEAD`?

philphauler:
Closing in favour of #1587 🙏 @sylvesterkaczmarek was right that the `#ifdef PTHREAD_MUTEX_ROBUST` guard is never true on glibc (enum constant, not a macro), so this branch never made the mutex robust on Linux. #1587 makes it a config option with portable robust / no-robust variants (no `#ifdef`, which also answers @joelsherrill on RTEMS/FACE), fixes the workflow failures @dzbaker flagged, and adds a functional test that deletes the owner and takes the mutex. Sorry for the delay.

Issue #1560, opened 2026-07-14
Add QNX task affinity support


dmolock-NASA:
Add task affinity support for the QNX operating system to support SMP operations.

PR #1574, opened 2026-09-04
Issue #1560, Fix Add QNX task affinity support

dmolock-NASA:
---
name: FSW Code Change
about: Flight Software code changes
labels: fsw
---

## Linked Issue

Closes #

## Requirements Impact

- Requirement ID(s):
- [ ] Requirements updated as necessary
- [ ] Existing requirements are still satisfied by this change

## Areas of Expertise Touched

- [ ] ASTRO
- [ ] CI/CD
- [ ] COSMOS
- [ ] Cybersecurity
- [ ] Docker
- [ ] EDS
- [ ] Git
- [ ] PSPs
- [ ] SBN
- [ ] SMP
- [ ] Tables
- [ ] TSN
- [ ] Unit Tests
- [ ] Other

---

## Author Checklist

- [ ] Linked GitHub issue is referenced above
- [ ] Code has been formatted with `.clang-format`
- [ ] Static analysis workflows ran and passed
- [ ] Unit tests (UT Assert) updated/added to cover code changes
- [ ] Unit test workflows ran and passed
- [ ] COSMOS test suite was run; tests updated/added if relevant changes were made
- [ ] Requirements have been reviewed; updated or confirmed still satisfied (see above)
- [ ] Testing evidence is included above
- [ ] Self-review of the diff completed

---

## Reviewer Checklist

- [ ] Code logic is correct and matches the stated intent
- [ ] Code is readable, maintainable, and follows project conventions (ask your lead if you are unsure of where to find these conventions)
- [ ] `.clang-format` has been applied
- [ ] Static analysis results reviewed and acceptable
- [ ] **The change has been exercised by the unit tests** (not just that tests pass — the new/changed code paths are actually covered)
- [ ] **COSMOS test suite was executed against this change** and results reviewed (or confirmed N/A with justification)
- [ ] **Reviewer has independently verified the change behaves as described** (e.g., by running the tests locally, reviewing CI output in detail, or performing additional ad-hoc testing as warranted)
- [ ] **Memory safety** reviewed (allocation, bounds, lifetime, stack usage)
- [ ] Requirements impact reviewed and appropriate
- [ ] Error handling is appropriate
- [ ] Appropriate Expert areas have been reviewed

### DIFF
--- a/src/os/qnx/CMakeLists.txt
+++ b/src/os/qnx/CMakeLists.txt
@@ -22,7 +22,6 @@ set(POSIX_BASE_SRCLIST
     ../posix/src/os-impl-queues.c
     ../posix/src/os-impl-tasks.c
     ../posix/src/os-impl-timebase.c
-    ../portable/os-impl-pthread-no-affinity-np.c
 )
 
 # Use portable blocks for basic I/O
@@ -98,6 +97,7 @@ endif ()
 set(QNX_IMPL_SRCLIST
     src/os-impl-filesys.c
     ../portable/os-impl-no-network-gethostid.c    # non-implemented versions of gethostid  API
+    src/os-impl-task-affinity.c
 )
 
 # Defines an OBJECT target named "osal_qnx_impl" with selected source files
--- a/src/os/qnx/inc/os-impl-qnx-task-affinity.h
+++ b/src/os/qnx/inc/os-impl-qnx-task-affinity.h
@@ -1,0 +1,260 @@
+/************************************************************************
+ * NASA Docket No. GSC-19,200-1, and identified as "cFS Draco"
+ *
+ * Copyright (c) 2023 United States Government as represented by the
+ * Administrator of the National Aeronautics and Space Administration.
+ * All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain
+ * a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ************************************************************************/
+
+/**
+ * \file
+ *
+ * \ingroup  posix
+ *
+ */
+
+#ifndef OS_IMPL_QNX_TASK_AFFINITY_H
+#define OS_IMPL_QNX_TASK_AFFINITY_H
+
+#include <string.h>
+#include <sys/syspage.h>
+#include <sys/neutrino.h>
+
+#include "osconfig.h"
+#include "common_types.h"
+#include "osapi-printf.h"
+#include "os-posix.h"
+
+#define OS_QNX_MAX_CPUS 128
+
+/**
+ * @brief QNX cpuset affinity structure
+ *
+ * This is used to represent the cpuset/affinity required by the QNX
+ * operating system (struct _thread_runmask) to set the runmask and inherit mask.
+ *
+ * Applications should not directly access fields within this structure,
+ * as the pointer configuration may change based on the number of processors in
+ * the running system.
+ */
+typedef struct
+{
+    int32  size; /**< The size of the QNX run/inherit mask */
+    uint32 affinity_masks[((OS_QNX_MAX_CPUS + 31) / 32)
+                          * 2]; /**< The QNX run and inherit masks storage (supports maximum number of processors) */
+} OS_QNX_cpuset_t;
+
+/*-------------------------------------------------------------------------------------*/
+/**
+ * @brief Initialize and zero the cpuset/affinity structure
+ *
+ * @param[in]  cpuset The cpuset/affinity structure
+ *
+ */
+static inline void OS_QNX_CpusetZero(OS_QNX_cpuset_t *cpuset)
+{
+    int32 MaskSize = RMSK_SIZE(_syspage_ptr->num_cpu);
+
+    /* Initialize/zero cpuset structure */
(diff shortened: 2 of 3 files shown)


src/os/qnx/CMakeLists.txt
arodarrow: I think we should add logic similar to what we have in other OS to check if (OSAL_CONFIG_MAX_CPUS GREATER 1)


if (OSAL_CONFIG_MAX_CPUS GREATER 1)
    list(APPEND QNX_IMPL_SRCLIST
        src/os-impl-task-affinity.c #    use the real affinity code
    )
else()
    list(APPEND QNX_IMPL_SRCLIST
        ../portable/os-impl-pthread-no-affinity-np.c
    )

src/os/qnx/inc/os-impl-qnx-task-affinity.h
```diff
@@ -0,0 +1,260 @@
...
+#include <string.h>
+#include <sys/syspage.h>
+#include <sys/neutrino.h>
+
+#include "osconfig.h"
+#include "common_types.h"
+#include "osapi-printf.h"
+#include "os-posix.h"
+
+#define OS_QNX_MAX_CPUS 128
```
arodarrow: Do we need to have this hardcoded number here? Is it possible to use OS_MAX_CPUS instead similar to the way it is done for OS_cpuset_t?

jphickey, line 38: This is a big number.  Are we really considering boards with 128 CPUs?  I would think 8 would be a more reasonable starting place.

Also - why not use OS_CONFIG_MAX_CPUS?

src/os/qnx/inc/os-impl-qnx-task-affinity.h
```diff
@@ -0,0 +1,260 @@
...
+
+/*-------------------------------------------------------------------------------------*/
+/**
+ * @brief Initialize and zero the cpuset/affinity structure
+ *
+ * @param[in]  cpuset The cpuset/affinity structure
+ *
+ */
+static inline void OS_QNX_CpusetZero(OS_QNX_cpuset_t *cpuset)
+{
```
jphickey, line 65: Why are all these declared as `static inline` in the header file?

They are clearly not referenced anywhere outside of the unit (and I would hope they are not, given the specificity to however QNX implements this) so my recommendation would be to put them in the C file, not the H file.

We should only put code in H files when there is a real justified reason to do it that way, but I'm not seeing that here.

src/os/qnx/inc/os-impl-qnx-task-affinity.h
```diff
@@ -0,0 +1,260 @@
...
+
+    /* Initialize/zero cpuset structure */
+    memset(cpuset, 0x0, sizeof(OS_QNX_cpuset_t));
+
+    /* Set size of run mask and inherit mask */
+    cpuset->size = MaskSize;
+
+    /*  Verify that cpuset has enought storage for cpus in system */
+    if ((sizeof(int32) + (sizeof(uint32) * MaskSize * 2)) > sizeof(OS_QNX_cpuset_t))
+    {
```
jphickey, line 76: this logic could use some explanation.

src/os/qnx/src/os-impl-task-affinity.c
```diff
@@ -0,0 +1,167 @@
...
+    int                             return_value;
+    uint32                          i;
+
+    impl = OS_OBJECT_TABLE_GET(OS_impl_task_table, *token);
+
+    /* Intitialize and zero cpuset */
+    OS_QNX_CpusetZero(&lcl_cpuset);
+
+    /* Populate POSIX cpuset from OSAL CPU Set using the macro */
+    for (i = 0; (i < OS_MAX_CPUS) && (i < OS_TaskAffinityGetCoresConfigured()); i++)
```
jphickey, line 97: Note that `OS_TaskAffinityGetCoresConfigured()` is going to be invoked for every iteration of the loop. 

It would be better to sample it once at the start into a local var and use that.

src/os/qnx/src/os-impl-task-affinity.c
```diff
@@ -0,0 +1,167 @@
...
+        {
+            /* If the bit is set in the POSIX structure */
+            if (OS_QNX_CpusetIsSetCore(&local_cpuset, i))
+            {
+                /* set the corresponding bit in the OSAL structure */
+                OS_CPUSET_SET(i, cpuset);
+            }
+        }
+
+        return OS_SUCCESS;
```
jphickey, line 159: Preference is to keep a local status variable and assign it, rather than a `return` in an if statement like this.

src/os/qnx/src/os-impl-task-affinity.c
```diff
@@ -0,0 +1,167 @@
...
+
+    impl = OS_OBJECT_TABLE_GET(OS_impl_task_table, *token);
+
+    /* Intitialize and zero cpuset */
+    OS_QNX_CpusetZero(&local_cpuset);
+
+    /* Get the runmask and inherit mask when passing cpuset set to zero */
+    return_value = ThreadCtlExt(0, (int)impl->id, _NTO_TCTL_RUNMASK_GET_AND_SET_INHERIT, (void *)&local_cpuset);
+
+    if (return_value != -1)
```
jphickey, line 143: Was expecting a check for 0 (pr maybe a positive number).
Does the QNX API _guarantee_ that the return value of `ThreadCtlExt()` is -1 (and only -1) on error, and every other possible value means success?

jphickey (changes requested):

src/os/qnx/src/os-impl-task-affinity.c
dmolock-NASA: Yes, the QNX API guarantees a return value of -1 on error. From the documentation: If an error occurs, these functions return -1 and set [errno](https://www.qnx.com/developers/docs/8.0/com.qnx.doc.neutrino.lib_ref/topic/e/errno.html). Any other value returned indicates success.

PR #1576, opened 2026-09-10
Fix #1553, guard POSIX state-change waits against spurious wakeups

### LINKED ISSUE #1553: Safeguard pthread conditional waits from spurious wake-ups.
Several instances of pthread_cond_wait or _timedwait without bounds this can cause concurrency problems if multiple threads are sharing condition variable.

For this issue verify that all threads are using separate, unique condition variables, and are safeguarded against spurious wake-ups with a loop. Example for one below:

do {
wait_result = pthread_cond_timedwait(&impl->cond, &impl->mutex, &ts);

    if (OS_impl_objtype_state_changed(objtype))
    {
        break;
    }
} while (wait_result == 0);

sylvesterkaczmarek:
Fixes #1553.

## Contribution

Track object-table notifications under the existing POSIX mutex and keep waiting through spurious condition-variable wakeups. Reuse the original absolute deadline, stop on timeout/error, and preserve cancellation cleanup.

The binary-semaphore wait already checks its value/flush predicate. Public condition-variable waits retain their documented caller-owned predicate semantics; no public API changes.

## Testing

Debian 12 arm64, GCC 12.2, CMake 3.25:
- All 60 coverage targets passed, including seven new POSIX idmap cases with 34 assertions.
- The final regression tests fail against unchanged upstream implementation in an isolated build, then pass with the fix.
- Full CTest run: 75/86 passed. The same 11 functional-test failures/timeouts also occur on unchanged upstream in this container; the baseline additionally fails the new regression target.
- Changed C/C++ lines formatted with the cFS configuration; `git diff --check` passed.

Tests cover repeated spurious wakes, actual notifications, unrelated object types, timeout/error exits, fixed deadlines, mutex isolation, and cleanup/initialization paths.

## Checklist

- [x] Contributing guide reviewed.
- [x] Issue and existing pull requests checked for overlap.
- [x] Regression coverage added and exercised.
- [ ] Contributor License Agreement submission verified.

Contributor: Sylvester Kaczmarek, Personal.
No third-party code added.

### DIFF
--- a/src/os/posix/src/os-impl-idmap.c
+++ b/src/os/posix/src/os-impl-idmap.c
@@ -112,6 +112,8 @@ void OS_Unlock_Global_Impl(osal_objtype_t idtype)
 
     if (impl != NULL)
     {
+        ++impl->change_count;
+
         /* Notify any waiting threads that the state _may_ have changed */
         ret = pthread_cond_broadcast(&impl->cond);
         if (ret != 0)
@@ -138,6 +140,8 @@ void OS_WaitForStateChange_Impl(osal_objtype_t objtype, uint32 attempts)
 {
     OS_impl_objtype_lock_t *impl;
     struct timespec         ts;
+    uint32                  change_count;
+    int                     wait_status;
 
     impl = OS_impl_objtype_lock_table[objtype];
 
@@ -166,7 +170,14 @@ void OS_WaitForStateChange_Impl(osal_objtype_t objtype, uint32 attempts)
         ++ts.tv_sec;
     }
 
-    pthread_cond_timedwait(&impl->cond, &impl->mutex, &ts);
+    /* A wake-up alone is not evidence that a table owner released the lock.
+     * Keep the original deadline across spurious wakes so the retry backoff
+     * neither expires early nor grows indefinitely. */
+    change_count = impl->change_count;
+    do
+    {
+        wait_status = pthread_cond_timedwait(&impl->cond, &impl->mutex, &ts);
+    } while (wait_status == 0 && impl->change_count == change_count);
 
     pthread_cleanup_pop(false);
 }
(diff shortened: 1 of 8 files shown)


src/os/posix/src/os-impl-idmap.c
```diff
@@ -166,7 +170,14 @@ void OS_WaitForStateChange_Impl(osal_objtype_t objtype, uint32 attempts)
...
 
-    pthread_cond_timedwait(&impl->cond, &impl->mutex, &ts);
+    /* A wake-up alone is not evidence that a table owner released the lock.
+     * Keep the original deadline across spurious wakes so the retry backoff
+     * neither expires early nor grows indefinitely. */
+    change_count = impl->change_count;
+    do
+    {
+        wait_status = pthread_cond_timedwait(&impl->cond, &impl->mutex, &ts);
+    } while (wait_status == 0 && impl->change_count == change_count);
```
jphickey, line 180: This should not be necessary as the calling thread already checks for the condition its looking for.  Doesn't matter why the task got woken up (spurious or not).
The only thing that matters is that the mutex gets re-acquired before pthread_cond_timedwait returns.
You may be correct that we should be looking at the status value returned by pthread_cond_timedwait .... there are a couple errno values that suggest a non-recoverable error, but the loop count is limited already so it will already give up if something like that happened.

sylvesterkaczmarek: Agreed. I traced the shared caller again: after every `OS_WaitForStateChange_Impl()` return it re-enters the bounded acquisition loop and rechecks the actual object state/refcount under the table lock. A spurious wake therefore does not bypass the condition being waited on, and the added per-table change counter is redundant. I’m withdrawing this PR rather than adding state that OSAL does not need.

PR #1587, opened 2026-09-17
Fix nasa/cFE#2433, make POSIX mutexes robust so a deleted owner cannot deadlock the takers

philphauler:
**Describe the contribution**

Supersedes #1573, same fix done properly. Fixes nasa/cFE#2433.

During shutdown every task is cancelled; a task cancelled at a cancellation point while holding a mutex leaves it locked forever and every task that later pends on it deadlocks. POSIX.1-2008 robust mutexes are the fix: the next taker gets the lock with `EOWNERDEAD` and restores it with `pthread_mutex_consistent()`.

What changed vs #1573, and why 👇

- @sylvesterkaczmarek was right: `#ifdef PTHREAD_MUTEX_ROBUST` is never true on glibc, the symbol is an enum constant, so #1573 compiled the robustness out and left the deadlock in place while calling an undeclared `pthread_mutex_consistent()` (the MCDC build failure). Verified with a two-line compile test against `/usr/include/pthread.h`.
- @joelsherrill's point about RTEMS and FACE: the feature is now an option, `OSAL_CONFIG_POSIX_ROBUST_MUTEX` (default TRUE), implemented as two portable source variants selected in CMake, the same pattern the posix layer already uses for CPU affinity. `os-impl-posix-mutex-robust.c` uses the 2008 API and is compiled with `_GNU_SOURCE` so glibc declares it (the BSP's `_XOPEN_SOURCE=600` does not); `os-impl-posix-mutex-no-robust.c` is a no-op. No `#ifdef` in the mutex code, which also clears the two CodeQL conditional-compilation alerts.
- @dzbaker: workflows addressed. Commit message follows the cFS pattern, clang-format applied with the cFS dev config, the stray BOM from #1573 is gone, no implicit declarations.
- A functional test, `mutex-robust-test`: a task takes a mutex, is deleted while holding it, and the next `OS_MutSemTake()` must succeed and the mutex must stay usable. Only built for the posix OS type with the option on.

**Testing performed**

Ubuntu 24.04 (WSL2, kernel 6.6), generic-linux BSP, `-DENABLE_UNIT_TESTS=TRUE`, run as root (SCHED_FIFO):
- with the change: `mutex-robust-test` 9/9, full `ctest` 86/86 (functional + coverage).
- with `-DOSAL_CONFIG_POSIX_ROBUST_MUTEX=FALSE` and the test forced in: `OS_MutSemTake` after `OS_TaskDelete` never returns; both threads sit in `futex_wait_queue`, the process ignores SIGTERM for 39 minutes (the shutdown path deadlocks on the same mutex) and only SIGKILL ends it. That is the #2433 behaviour.

**Expected behavior changes**

None on platforms with the option off. With it on (default), a mutex whose owner was deleted is recoverable by the next taker instead of deadlocking it.

**System(s) tested on**

Ubuntu 24.04 / glibc 2.39 / GCC 13.3 under WSL2.

**Additional context**

RTEMS and FACE profiles should set the option FALSE; happy to wire an auto-detect (`check_symbol_exists(pthread_mutex_consistent)`) if you would rather not have a manual switch.

**Contributor Info**

Filip Koscak (philphauler). AI used: yes, Claude Code for the analysis, the variant split and the test; reviewed by hand, the proof runs are real output.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01HDRvC9ePDP3i6xpxFAECFg


sylvesterkaczmarek (approved):
Re-reviewed this as the replacement for #1573. The Linux robustness issue I raised there is addressed properly here: the implementation no longer relies on `#ifdef PTHREAD_MUTEX_ROBUST`, the robust API is selected explicitly in CMake, `EOWNERDEAD` is recovered with `pthread_mutex_consistent()`, and there is a functional owner-death regression. The full OSAL CI, including MCDC, format, CodeQL, and tests, is green on this head. No blocker from me.

philphauler:
🚀🚀🚀🚀

